PR #7, opened 2017-11-08
Move grid-specific functionality to grid folder and add tests

jenniferarnesen:
"Magic" numbers related to the grid have all been moved into gridUtil, instead of being split between the GridItem component and the old util module.


turban (approved):
Looks fine to me :-)

PR #24, opened 2018-01-09
Add ability to delete your own interpretation

jenniferarnesen:


edoardo (approved):
Seems good.
The only comment I have is about measure units in the width/height CSS rules.
Probably not required and the default is px, but it would be nice to find out what's the best practice.

PR #25, opened 2018-01-11
Add initial style guide

jenniferarnesen:
I was thinking that we can store the color codes in one place. If we agree on this concept, it might be that it gets moved to d2-ui, or its own repo.


jenniferarnesen:
Will address common styles/theme later

PR #34, opened 2018-01-17
Implement Messages grid item

jenniferarnesen:
Includes:
- add baseUrl to the App context so that sub-components can form absolute href urls
- separate out the ItemHeader component so different types of Items can use it
- moved some date functions to util so they can be used by multiple components/modules
- messages property on the redux store root


edoardo (approved):
One comment about style handling. Do we prefer a .css file associated to the component or defining styles in the code?

It would be nice to integrate push notifications in this component later on 🙂  (https://github.com/edoardo/dhis2-message-conversations-notifier-app)

PR #47, opened 2018-01-22
Separate components for View and Edit Titlebar

jenniferarnesen:
This is to keep the TitleBar component from getting too complex with both Edit and View functionality.


jenniferarnesen:
@janhenrikoverland I tried to simplify things even more by moving the code that figures out the name and description, into the TitleBar component, but I ran into problems. Maybe we can take a look at it together?

PR #57, opened 2018-01-24
Add AppItem dashboard item

edoardo:


edoardo:
Note: when testing on port :3000 the iframe won't load the app due to the `X-Frame-Options` header set to `sameorigin`.

jenniferarnesen (approved):

PR #58, opened 2018-01-25
Save dashboard items

jenniferarnesen:
Regarding width and height (vs w and h), I decided to go for the least intrusive solution, which was just to change the name of those props when GETing and PUTing to the api. Since the use of w and h (instead of width and height) use used in several places in the app.

### DIFF
--- a/src/actions/editDashboard.js
+++ b/src/actions/editDashboard.js
@@ -27,6 +27,11 @@ export const acSetDashboardDescription = value => ({
     value,
 });
 
+export const acUpdateDashboardLayout = value => ({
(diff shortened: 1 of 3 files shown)


edoardo (approved):
2 comments:
1) does it save much work converting w/h back and forth?
2) is d2.models save not working after assigning name, description and dashboardItems?

jenniferarnesen:
@edoardo The react-grid-layout expects w and h, so we have to convert at some point. Regarding d2, I had tested previously (before the api changes) trying to supply dashboardItems, but d2 seemed to simply extract the id's and only pass those. I can try again though, before merging.

PR #60, opened 2018-01-25
Move List item content delete icon next to resource name

jenniferarnesen:
The reason for moving the icon to be just to the right of the resource name is so that it isn't so close to the delete icon for the entire item. Also, made it smaller to further distinguish it from the Item delete button. A horizontal line was also added to separate the header from the content area of the item, similar to the messages type.

Some refactoring of all the different ways we are currently doing icon "buttons" is needed.


edoardo (approved):
I'm using the Divider MUI component in the ItemSelectList.
Not sure how is different from the Line one, maybe using a MUI component for something that simple is a bit overkill.
I agree about the buttons, deciding on a unique way of using them makes sense.

PR #63, opened 2018-01-26
Add button that links to visualization, and show description for plugin items

edoardo:


jenniferarnesen:
After discussion with Joe about the complexities of having 2 buttons that affect the expanded item area (description and interpretations), we agreed to get rid of the description button. The description should always show with the interpretations anyway, so the "interpretations" button will now also be the way for users to see the description.

PR #64, opened 2018-01-26
Feature/share

edoardo:
It's missing some translations for the strings.
I still find strange some string translations for components must be defined in the app... same problem in the new map app.
Will see if there is a way for the SharingDialog component to have all the strings.
The whole i18n system is still obscure for me :wink:


edoardo:
Added a properties file with the missing translations for the sharing dialog.
To make it work in dev mode running localhost:3000, copy the whole src/i18n folder to public.

PR #65, opened 2018-01-26
Configure baseurl and authorization for development

jenniferarnesen:
* Remove credentials and hardcoded urls from the code and rely on the user's DHIS2_HOME config instead. 
* Removed all the verbose comments from the webpack.config files, that documentation can be found in the create-react-app git repo.
* Use the webpack module devserver property to configure proxying for resources, instead of a hardcoded config in package.json


jenniferarnesen:
Breaking this up into smaller chunks

PR #78, opened 2018-02-01
Enable switch to map button, fix interpretations handling

edoardo:
- Use the getLink in plugin.js everywhere
- Fix interpretations handling for all plugin item types
- Enable switch to map button in plugin item header
- Point map URL to the new app


src/Item/PluginItem/ItemFooter.js
```diff
@@ -54,6 +54,7 @@ class ItemFooter extends Component {
                     <ItemDescription description={description} />
                     <hr style={style.line} />
                     <Interpretations
+                        object={this.props.item}
```
jenniferarnesen: What about just sending object, and not objectType and objectId, since Interpretations can derive those values from object? Seems like a lot of noise to break apart object and send all its different properties separately

edoardo: Removed objectType, but kept objectId to avoid having to call getId everywhere.

jenniferarnesen (approved):

PR #81, opened 2018-02-05
Configure d2 to use a specific api version

jenniferarnesen:
Set the version in the manifest, which is configured in package.json. Webpack then builds the manifest information into the bundle. Then, instead of asking d2 to read the manifest, the app will read the manifest itself and then configure d2 with the correct api url.

### DIFF
--- a/config/webpack.config.prod.js
+++ b/config/webpack.config.prod.js
@@ -27,7 +27,7 @@ const shouldUseSourceMap = process.env.GENERATE_SOURCEMAP !== 'false';
 const publicUrl = publicPath.slice(0, -1);
 // Get environment variables to inject into our app.
 const env = getClientEnvironment(publicUrl);
-const manifest = require(`${paths.appPublic}/manifest`);
+const manifest = require(`${paths.appBuild}/manifest`);
 const globals = Object.assign(
     {},
     {
@@ -35,6 +35,10 @@ const globals = Object.assign(
     },
     env.stringified
 );
+
+console.log('paths', paths);
+
+console.log('manifest', manifest);
 
 // Assert this just to be safe.
 // Development builds of React are slow and not intended for production.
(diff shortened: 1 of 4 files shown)


config/webpack.config.prod.js
```diff
@@ -27,6 +27,14 @@ const shouldUseSourceMap = process.env.GENERATE_SOURCEMAP !== 'false';
 const publicUrl = publicPath.slice(0, -1);
 // Get environment variables to inject into our app.
 const env = getClientEnvironment(publicUrl);
+const manifest = require(`${paths.appBuild}/manifest`);
+const globals = Object.assign(
+    {},
+    {
+        manifest: JSON.stringify(manifest),
+    },
+    env.stringified
```
varl, line 39: Is this assigning a string to an `Object`? The name `env.stringified` implies so. That seems strange to me so I'm curious how this looks at runtime. When playing around with it I got this:

```
Object.assign({foo:'bar'}, {foo:'buz', bar: 'foo'}, 'hello,bar,foo')
// returns: { '0': 'h',
//  '1': 'e',
//  '2': 'l',
//  '3': 'l',
//  '4': 'o',
//  '5': ',',
//  '6': 'b',
//  '7': 'a',
//  '8': 'r',
//  '9': ',',
//  '10': 'f',
//  '11': 'o',
//  '12': 'o',
//  foo: 'buz',
//  bar: 'foo' }

```

So if `env.stringified` is in fact a string maybe it should be `env` to get the properties added to the new object? Am I missing something here?

janhenrikoverland (approved):

config/webpack.config.prod.js
jenniferarnesen: Its actually a confusing prop name. env.stringified is an object.

PR #86, opened 2018-02-06
Implement spacer items for creating empty space between dashboard items

jenniferarnesen:
The spacer item is a client-side-only type. On the server side, it is a Text item type with a specific text value.

### DIFF
--- a/src/Item/SpacerItem/Item.js
+++ b/src/Item/SpacerItem/Item.js
@@ -1,0 +1,22 @@
+import React, { Fragment } from 'react';
+import ItemHeader from '../ItemHeader';
+
+const style = {
+    margin: '10px',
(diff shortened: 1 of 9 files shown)


src/Item/SpacerItem/Item.js
```diff
@@ -0,0 +1,22 @@
...
+    fontSize: '15px',
+    lineHeight: '18px',
+};
+
+const SpacerItem = () => {
+    return (
+        <Fragment>
+            <ItemHeader title="Spacer" />
+            <p style={style}>
+                Use a spacer to create empty vertical space between other
```
edoardo, line 15: Is it always only vertical?

edoardo (approved):

PR #90, opened 2018-02-07
Add delete access check for deleting interpretations and comments

jenniferarnesen:
According to the backend team, the delete access check is as follows:

To delete an interpretation: user owns the interpretation or has delete access. The result of deleting an interpretation is that all the comments are deleted.
To delete a comment from an interpretation: user owns the comment, or user is a superuser.

Note the inconsistent behavior: the owner of the interpretation can delete a comment if they delete their whole interpretation. However he/she cannot delete individual comments.

### DIFF
--- a/src/Item/PluginItem/Interpretations/Interpretation.js
+++ b/src/Item/PluginItem/Interpretations/Interpretation.js
@@ -141,13 +141,12 @@ class Interpretation extends Component {
     };
 
     userIsOwner = id => id === this.context.d2.currentUser.id;
+    canDelete = userId =>
+        this.userIsOwner(userId) || this.props.interpretation.access.delete;
 
     renderActions() {
         const likes =
             this.props.interpretation.likedBy.length === 1 ? 'like' : 'likes';
-        const userOwnsInterpretation = this.userIsOwner(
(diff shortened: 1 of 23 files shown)


src/Item/PluginItem/Interpretations/Interpretation.js
```diff
@@ -252,21 +258,24 @@ class Interpretation extends Component {
...
-        likeInterpretation: data => dispatch(tLikeInterpretation(data)),
-        unlikeInterpretation: data => dispatch(tUnlikeInterpretation(data)),
-        addComment: data => dispatch(tAddInterpretationComment(data)),
-        deleteComment: data => dispatch(tDeleteInterpretationComment(data)),
-        deleteInterpretation: data => dispatch(tDeleteInterpretation(data)),
-    };
-};
+const mapStateToProps = state => ({
+    userId: fromUser.sGetUserId(state),
+    isSuperUser: fromUser.sGetIsSuperuser(state),
```
edoardo: Maybe rename this to userIsSuperUser or something that refers to the user, similarly to userIsOwner above

edoardo (approved):

PR #93, opened 2018-02-08
Use common functions for icons/urls, add icons in ListItem

edoardo:


src/util.js
```diff
@@ -83,3 +84,9 @@ export const getBaseUrl = d2 => {
     const idx = api.baseUrl.indexOf('/api');
     return idx > -1 ? api.baseUrl.slice(0, idx) : api.baseUrl;
 };
+
+export const getItemUrl = (type, id, d2) => {
```
jenniferarnesen: Would it make more sense for this function to be in the itemTypes file?

jenniferarnesen (changes requested):

jenniferarnesen (approved):

src/util.js
edoardo: done

PR #95, opened 2018-02-08
Improve look of Text item in edit and view mode

jenniferarnesen:
- Make the view and edit mode look more alike with margins, padding, centering, font size, line height, so the user gets a better idea of how it will look in view mode.
- Add ItemHeader in Edit mode
- If the user saves a text item that has no content, then we need to save some placeholder text, because the backend requires the text property to have content. Convert this "empty string" content to "" for the UI


edoardo (approved):
Only one comment about the placeholder, remove the "No content" text and fix the placeholder position.

PR #98, opened 2018-02-09
Task/translation button

edoardo:
I noticed some errors in the console from the Translation component.
I think they trigger whenever there is an empty text field for name and/or description.
I don't think it's a problem of the app.


jenniferarnesen (approved):
Just pull out the TranslationDialog to a separate function to keep render() from getting too big.

PR #104, opened 2018-02-12
Rename the dashboard 'starred' property to 'favorite' when saving dashboard

jenniferarnesen:
The api expects a property named 'favorite', but in the frontend, 'starred' is a more meaningful and less dubious name (so keep the name 'starred' for frontend). This change also restricts the properties sent to the ones that the user may have changed. (except for the 'id' property, which is obviously needed for identification).

### DIFF
--- a/src/actions/editDashboard.js
+++ b/src/actions/editDashboard.js
@@ -95,8 +95,12 @@ export const tSaveDashboard = () => async (dispatch, getState) => {
         };
     });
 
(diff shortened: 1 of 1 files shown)


jenniferarnesen:
Note that the backend save api doesn't currently save the favorite property. I have informed Lars of this.

janhenrikoverland (approved):

jenniferarnesen:
This did not actually fix the bug. The fix has to be in d2.

PR #110, opened 2018-02-13
Disable or remove active links on items when in edit mode

jenniferarnesen:
So that the user doesn't inadvertently click themselves away from editing the dashboard.
- remove the Launch button on the items
- for messages items, set href=# so the messages look the same but don't do anything


edoardo:
What about links in ListItem?

janhenrikoverland (approved):

PR #111, opened 2018-02-14
i18next/basic integration

adeelshahid:
separate languages by **dhis2** namespace

- next is to write scripts to auto generate English translation keys
- merge mix with existing set of languages etc.


adeelshahid:
@jason-p-pickering example translation integration.

We will use some scripts to parse the code on pre-commit to generate the **en.js** file.

jason-p-pickering:
@adeelshahid we have created a document [here](https://docs.google.com/document/d/1I7t9dE46HUj4d7cieuub7eN6qa4DembN7wcYM_tNNGI/edit) which outlines the general approach we have used for integration of i18next with the other apps. The important parts are

1) POT files should be generated with the extractor script. This script may be slightly different for each app, but we have examples for both angular as well as react. This should occur during the pre-commit phase as you noted. This will allow the i18n of the app with a new template. Once we get the PO files back from the i18n workflow, these will be committed. 
2) During the build phase, all PO files will be transformed into an object(s) for integration into the built artifact. 

If you need more details, let me or Joao know via Slack! :+1:

joaoantunes87 (commented):
So far, I think it is good.

PR #114, opened 2018-02-15
Add dashboard item filters

janhenrikoverland:
- Renamed current store "filter" to "dashboardsFilter"
- Added redux stuff for the new prop "itemFilter"
- Moved and added plugin update logic from componentDidMount to componentWillReceiveProps in PluginItem to be able to handle both the initial load and the required reloads correctly


jenniferarnesen:
Is being merged into another branch, and will make a PR from that.

PR #115, opened 2018-02-15
Fix scrollbars bug (DHIS2-2834)

edoardo:


edoardo:
Completely forgot about item resize.
The fixed height approach cannot be used.
Working on a new solution.

jenniferarnesen:
This code doesn't fix the issue unfortunately. Closing so that it doesn't inadvertently get merged.

PR #116, opened 2018-02-15
Add dialog for choosing org units to filter the dashboard with

jenniferarnesen:
The dialog provides the org unit tree (potentially multiple roots) for user to select org units for filtering the dashboard. The org unit filter impacts plugin items only.


jenniferarnesen:
This code was merged in a separate PR

PR #119, opened 2018-02-15
Remove components and code that are not in use

jenniferarnesen:
The removed code is available on another branch called code-for-2.30, if we want to incorporate it into the next version. But not good to keep unused code in the master branch that we may or may not use in the future.


jenniferarnesen:
This is so old, creating a new one.

PR #122, opened 2018-02-18
Add underlines for title and description edit fields

jenniferarnesen:
UI changes according the the sketches


edoardo (approved):
TitleBar.css can be removed then?

PR #124, opened 2018-02-18
Style dashboard item chips according to design

jenniferarnesen:
The dashboard item chips do not use d2-ui for now since we do not want to impact other apps with the color changes.

### DIFF
--- a/src/ControlBarContainer/DashboardsBar.js
+++ b/src/ControlBarContainer/DashboardsBar.js
@@ -75,25 +75,6 @@ const DashboardsBar = ({
 
     const controlBarHeight = getOuterHeight(rowCount, true);
 
-    const dashboardChips = () => {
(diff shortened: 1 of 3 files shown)


src/ControlBarContainer/DashboardsBar.js
```diff
@@ -105,12 +105,9 @@ const DashboardsBar = ({
                 </div>
                 {dashboards.map(dashboard => (
                     <Chip
-                        key={dashboard.id}
```
edoardo, line 108: Is the key not needed anymore?

edoardo (approved):

src/ControlBarContainer/DashboardsBar.js
jenniferarnesen: No, since the dashboard chips are not list elements

PR #127, opened 2018-02-18
Set the maximum height of the ItemSelect popover to stay within visible area

jenniferarnesen:
A slight hack to calculate the max height of the ItemSelect popover

### DIFF
--- a/src/ItemSelect/ItemSelect.js
+++ b/src/ItemSelect/ItemSelect.js
@@ -148,7 +148,18 @@ class ItemSelect extends React.Component {
             .catch(console.error);
     };
 
+    getMaxPopoverHeight = () => {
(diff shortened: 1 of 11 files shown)


src/ItemSelect/ItemSelect.js
```diff
@@ -167,8 +178,7 @@ class ItemSelect extends React.Component {
                     autoCloseWhenOffScreen={false}
                     style={{
                         width: 700,
-                        height: 900,
-                        overflow: 'auto',
+                        height: popoverHeight,
```
edoardo: What about using "viewport height" measure unit instead?
Something like `80vh.`

edoardo (approved):

PR #138, opened 2018-02-20
Add snackbar that shows message about loading dashboard

jenniferarnesen:
If loading the dashboard takes more than 500ms, then show a snackbar message indicating that loading is happening.

The snackbar can be reused for communicating other messages as well. Because of this, I left in the commented out line checking for message type in SnackbarMessage.js, even though there is currently only one message type. Future message types will be: Dashboard item added/removed. Possibly also error messages.

You might notice that I am reusing the loadingDashboardMsg object rather than creating a new object for each message. This is because of some buggy behaviour that is described in the MUI snackbar documentation.

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -40,6 +40,12 @@ class App extends Component {
     };
 
     render() {
+        console.log(
+            'render App, with snackbar prop',
+            this.props.snackbarMessage,
+            !!this.props.snackbarMessage
+        );
+
         return (
             <div className="app-wrapper">
                 <HeaderBar />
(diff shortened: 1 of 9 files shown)


src/App.js
```diff
@@ -30,6 +34,10 @@ class App extends Component {
         };
     }
 
+    onCloseSnackbar = () => {
+        this.props.acSnackbarClosed();
```
edoardo: Almost all the actions we have are named with the format ac + action + thing.
I would rename this to _acCloseSnackbar_.

edoardo (approved):

PR #139, opened 2018-02-20
Switch from map to table/chart

janhenrikoverland:
- Use first thematic map view as config when switching to table or chart
- For all visualisation types: when switching back to the original type - just use the id and let the plugin handle it instead of de-transforming the transformed config - that request will be cached as well
- Restructured api functions and get-fields functions + adding a new file called metaData.js


jenniferarnesen:
@janhenrikoverland Looks good. Just 2 observations:
- It seems wrong the apiFetchFavorite is in metaData.js. Seems like that fn can't find a proper home. In which case maybe it should be in its own file? But its definitely doesn't have anything to do with metadata, right?
- To me, the camelcase of metaData seems weird. I think it would be better as metadata.js, since metadata has long since become a word in its own right, not just a compound word.

jenniferarnesen (approved):

PR #141, opened 2018-02-22
Store active visualisation

janhenrikoverland:
Update and listen to store.visualizations.<id>.activeType when switching between visualization types.


src/Item/PluginItem/Item.js
```diff
@@ -128,4 +185,17 @@ Item.contextTypes = {
...
-export default Item;
+const mapStateToProps = (state, ownProps) => ({
+    itemFilter: fromItemFilter.sGetFromState(state),
+    visualization: sGetVisualization(
+        state,
+        pluginManager.extractFavorite(ownProps.item).id
+    ),
+});
+
+const mapDispatchToProps = dispatch => ({
```
jenniferarnesen, line 219: I think its possible to write this in shorthand:

connect(mapStateToProps, {acReceivedActiveVisualization})(Item);

Then, in the component:

this.props.acReceivedActiveVisualization(id, type, activeType);

src/reducers/visualizations.js
```diff
@@ -1,29 +1,48 @@
+import { orObject } from '../util';
+import objectClean from 'd2-utilizr/lib/objectClean';
+
 /** @module reducers/visualizations */
 
+const isEmpty = p => p === undefined || p === null;
```
jenniferarnesen, line 6: Why not also use the d2-utilizr isEmpty function, as you did in the itemFilter reducer?

src/reducers/visualizations.js
```diff
@@ -1,29 +1,48 @@
...
-            const res = Object.assign({}, state, {
+            return {
+                ...state,
                 [action.value.id]: action.value,
-            });
-
-            return res;
+            };
+        }
+        case actionTypes.RECEIVED_ACTIVE_VISUALIZATION: {
```
jenniferarnesen, line 23: Please add a reducer unit test for the new action

jenniferarnesen (changes requested):
You'll need to merge master into this and probably make a few changes. See commit: a784010

src/Item/PluginItem/Item.js
janhenrikoverland: But then we give this presentational component a crazy api. Presentational components should be reusable ideally.

When I use a ui component and want to pass in a handler I would expect to pass in "onSelectX" or similar, not "acReceivedActiveVisualization".

We could of course use "onSelectVisualization" and still do the shorthand. I usually extract mapState and mapDispatch, but that's just for readability.

src/reducers/visualizations.js
janhenrikoverland: Because it considers empty array to be empty and thus removes "interpretations": [ ] from the visualization object.

src/reducers/visualizations.js
janhenrikoverland: Done

jenniferarnesen (approved):

PR #143, opened 2018-02-23
Fix crash if user has no permission to see analytics apps

edoardo:
If a user does not have permission to see an analytics app (in user
roles) access to the whole app folder is denied, which causes the plugin
request to fail, causing errors when trying to call a function on it (like
load/reload).
Instead, show a message informing the user about the unavailable plugin.
We could improve the message suggesting the likely reason for it:
missing permission.
The whole header is also removed, to avoid other errors when clicking
the buttons (ie. launch).
Switching between different visualisation types could work if a user can
see other visualisation apps, but it's inconsistent when trying to go
back to the original visualisation.
Even better would be to hide only the button related to the type of
visualisation that does not work.
Another solution, depending on what the "See app" permission really
means, is to make all the plugins always available, the user then can
see the dashboard item, but won't be able to use the associated app (no
launch button).

### DIFF
--- a/src/Item/PluginItem/Item.js
+++ b/src/Item/PluginItem/Item.js
@@ -9,6 +9,7 @@ import PluginItemHeaderButtons from './ItemHeaderButtons';
 import * as pluginManager from './plugin';
 import { getGridItemDomId } from '../../ItemGrid/gridUtil';
 import { getBaseUrl } from '../../util';
+import { itemTypeMap } from '../../itemTypes';
 
 const style = {
     icon: {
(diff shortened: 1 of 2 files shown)


src/Item/PluginItem/Item.js
```diff
@@ -36,22 +43,30 @@ class Item extends Component {
     state = {
         showFooter: false,
         activeVisualization: this.props.item.type,
+        pluginIsAvailable: itemTypeMap[this.props.item.type].plugin
```
jenniferarnesen: You can use the shorthand notation like this:

pluginIsAvailable: !!itemTypeMap[this.props.item.type].plugin

jenniferarnesen (approved):

PR #144, opened 2018-02-25
Save controlbar rows

janhenrikoverland:
- Fetch value from userDataStore on startup (there will be two 404s on the first load when the namespace does not yet exist for the user - harmless and can't be avoided)
- Save new value on end drag


src/api/controlBar.js
```diff
@@ -0,0 +1,35 @@
+import { getInstance } from 'd2/lib/d2';
+import { DEFAULT_ROWS } from '../reducers/controlBar';
+
+const NAMESPACE = 'dashboard';
```
jenniferarnesen, line 4: Since the code is very specifically about the Dashboard, and controlbar rows, it would make the code more readable to rename these constants to something more meaningful, like DASHBOARD (or NS_DASHBOARD, or NAMESPACE_DASHBOARD), and CONTROL_BAR_ROWS.

Alternatively, we could make this a more generic "userDataStore" file, and export different namespace and key constants for use in actions.

jenniferarnesen (approved):

PR #159, opened 2018-03-01
Use only row count to determine if dashboards bar is expanded

jenniferarnesen:
These changes fix https://jira.dhis2.org/browse/RP229-54

Changes included:
- follow the design and change the button titles to "View all dashboards" and "close".
- clicking on the view-all and close buttons will completely expand or contract the dashboards bar, not set it to something in between.
- the number of rows decides if the bar is at max height or not. No need for the "expanded" property.
- when dashboard bar is only partly expanded, then the button will say "view all dashboards" and clicking will expand to full height.


edoardo:
Seems to work well. Approved.

PR #165, opened 2018-03-05
Restore missing question mark in delete confirmation message

jenniferarnesen:
The dashboard name should be in bold to give it emphasis. In order to do this with internationalization, use the Trans component. The Trans component requires i18n in the context, so i18n was added at the root context (in App.js) so it will be available for any situations where Trans component is needed.

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -1,6 +1,7 @@
 import React, { Component } from 'react';
 import { connect } from 'react-redux';
 import PropTypes from 'prop-types';
+import i18n from 'dhis2-i18n';
 
 import HeaderBarComponent from 'd2-ui/lib/app-header/HeaderBar';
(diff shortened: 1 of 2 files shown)


edoardo (changes requested):
This need to change to use interpolation, otherwise the translators will get the wrong understanding of the actual string.
I'm not sure how you interpolate HTML tags but something like this is what we should use instead:

`i18n.t('Are you sure you want to delete dashboard {name}?', {name: '<strong>{dashboardName}</strong>', interpolation: { escapeValue: false }})`

edoardo (approved):

adeelshahid:
@edoardo @jenniferarnesen after release once #140 is merged we will only use 'd2-i18n' library for translations. I have removed *react-i18next*.

Translation .pot file generation etc. is automatically extracted out. Example documention: https://github.com/dhis2/d2-i18n

PR #184, opened 2018-04-18
Separate items from dashboards in redux store (DHIS2-3299)

janhenrikoverland:
- refactors state.dashboards into byId (the dashboards list) and items (array holding the current dashboard items) so that the grid could listen to changes to the items only and not dashboard related stuff like "starred" etc
- updated tests
- moved the page container top margin (which is listening to the number of controlbar row among other things) to a separate component so that the whole page (grid etc) does not reload when the controlbar height is changed
- changed the connected page container props from listening to the dashboards object reference to instead only check if the dashboards list is empty or null
- css fix that removes the extra bottom right corner arrow on items in edit mode (react-resizable)
- css update for the spacer item text in edit mode:
--- aligns the item content margin with the text item
--- makes the spacer description text a bit brighter so that it's clear that it's just a description and not actual text (like the text item)

### DIFF
--- a/src/Item/SpacerItem/Item.js
+++ b/src/Item/SpacerItem/Item.js
@@ -1,12 +1,13 @@
 import React, { Fragment } from 'react';
 import i18n from 'd2-i18n';
 import ItemHeader from '../ItemHeader';
+import { colors } from '../../colors';
 
 const style = {
     margin: '21px 28px',
     fontSize: '14px',
     lineHeight: '18px',
-    color: '#666',
+    color: colors.charcoalGrey,
 };
 
 const SpacerItem = () => {
--- a/src/Item/SpacerItem/Item.js
+++ b/src/Item/SpacerItem/Item.js
@@ -3,9 +3,10 @@ import i18n from 'd2-i18n';
 import ItemHeader from '../ItemHeader';
 
 const style = {
-    margin: '10px',
-    fontSize: '15px',
+    margin: '21px 28px',
+    fontSize: '14px',
     lineHeight: '18px',
+    color: '#666',
 };
 
 const SpacerItem = () => {
--- a/src/PageContainer/PageContainerSpacer.js
+++ b/src/PageContainer/PageContainerSpacer.js
@@ -1,0 +1,24 @@
+import React from 'react';
+import { connect } from 'react-redux';
+
+import { CONTROL_BAR_ROW_HEIGHT } from '../ControlBarContainer/ControlBarContainer';
+import { sGetIsEditing } from '../reducers/editDashboard';
+import { sGetControlBarUserRows } from '../reducers/controlBar';
+
+const DEFAULT_TOP_MARGIN = 80;
+
+const PageContainerSpacer = props => (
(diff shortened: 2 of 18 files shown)


src/reducers/__tests__/dashboards.spec.js
```diff
@@ -1,67 +1,131 @@
...
+                    ...currentBoards.dash1,
+                    starred: true,
+                },
+            },
+        };
+
+        expect(actualState).toEqual(expectedState);
+    });
+
+    it('SET_DASHBOARD_STARRED: should set "displayName" on a dashboard', () => {
```
jenniferarnesen: wrong action in test title

src/actions/editDashboard.js
```diff
@@ -19,10 +19,25 @@ const onError = error => {
...
+        dashboardItems: items,
+    };
+
+    return {
+        type: actionTypes.RECEIVED_EDIT_DASHBOARD,
+        value: val,
+    };
+};
+
+// export const acSetEditDashboard = (dashboard, items) => ({
```
jenniferarnesen: Any reason to keep this commented out code?

jenniferarnesen (changes requested):

jenniferarnesen:
I still think its not very intuitive to have dashboardItems inside the dashboards prop. But I'm not going to demand it.

Could you add a good title to the PR, suffixed with DHIS2-3299 so that the PR is linked to the jira issue?

In the description, could you go into a little detail about what has been done, including the purpose of the styling change to the spacer item?

src/Item/SpacerItem/Item.js
```diff
@@ -3,9 +3,10 @@ import i18n from 'd2-i18n';
 import ItemHeader from '../ItemHeader';
 
 const style = {
-    margin: '10px',
-    fontSize: '15px',
+    margin: '21px 28px',
+    fontSize: '14px',
     lineHeight: '18px',
+    color: '#666',
```
jenniferarnesen: Are you sure this color cant be approximated by one of the colors in colors.js (which are defined by Joe's UI color scheme)?

jenniferarnesen (changes requested):

jenniferarnesen:
Would PageContainerTopMargin be a more explanatory component name than PageContainerSpacer?

src/PageContainer/PageContainerSpacer.js
```diff
@@ -0,0 +1,24 @@
...
+import { connect } from 'react-redux';
+
+import { CONTROL_BAR_ROW_HEIGHT } from '../ControlBarContainer/ControlBarContainer';
+import { sGetIsEditing } from '../reducers/editDashboard';
+import { sGetControlBarUserRows } from '../reducers/controlBar';
+
+const DEFAULT_TOP_MARGIN = 80;
+
+const PageContainerSpacer = props => (
+    <div className="page-spacer" style={{ marginTop: props.marginTop }} />
```
jenniferarnesen: Is the class name needed?

jenniferarnesen (approved):

PR #187, opened 2018-04-26
Update dashboard selectors

janhenrikoverland:
- Add "dashboard" to selector names to make things more clear when imported directly
- Make sure get-by-id no longer returns all dashboards if no id is provided (or the id does not get a match)
- Add get-all-dashboards selector instead
- Give some selectors a more descriptive name
- Add selector tests


src/reducers/dashboards.js
```diff
@@ -117,31 +125,32 @@ export const sGetById = (state, id) =>
...
- * @param {Object} propName The name of the filter prop
- * @param {Object} value The value of the filter prop
+ * @param {String} propName The name of the filter prop
+ * @param {Any} value The value of the filter prop
  * @returns {Array}
  */
-const sGetDashboardsByProp = (state, propName, value) =>
-    Object.values(orObject(sGetById(state))).filter(
+const sGetDashboardsFilteredByPropValue = (state, propName, value) =>
+    Object.values(orObject(sGetAllDashboards(state))).filter(
```
jenniferarnesen: I don't think the orObject is necessary here since it is in sGetAllDashboards

src/reducers/dashboards.js
```diff
@@ -117,31 +125,32 @@ export const sGetById = (state, id) =>
...
  * @param {Object} state The current state
- * @param {Object} propName The name of the filter prop
- * @param {Object} value The value of the filter prop
+ * @param {String} propName The name of the filter prop
+ * @param {Any} value The value of the filter prop
  * @returns {Array}
  */
-const sGetDashboardsByProp = (state, propName, value) =>
-    Object.values(orObject(sGetById(state))).filter(
+const sGetDashboardsFilteredByPropValue = (state, propName, value) =>
```
jenniferarnesen: Should we really keep this function? It is not exported/used anywhere else in the app. Kind of confusing at first bc of the selector-like name and the docblock (which I know you didn't write...)

Instead the functions below that use it could just write the compact code like:

Object.values(sGetAllDashboards(state)).filter(d => d.starred === true);

jenniferarnesen (changes requested):
Could you add some tests for the selectors that were changed in this PR?

src/reducers/dashboards.js
janhenrikoverland: Correct

src/reducers/dashboards.js
janhenrikoverland: The idea is just that there might be a bunch of those coming later? So to avoid repeating "Object.values(sGetAllDashboards(state)).filter ...etc)" there is a generic function doing this instead.

janhenrikoverland: Not important, we could do that later if needed instead.

jenniferarnesen: Yeah, lets do it when we find that we need it.

jenniferarnesen (approved):

PR #191, opened 2018-06-05
Feature/edit interpretations [DHIS2-3430]

joakimia:
This feature enables editing of interpretations, and interpretation comments.
Two more actions are introduced to the store:
- tEditInterpretation
- tEditInterpretationComment

Two more props are added for the <InputField /> used for generating interpretations or comments:
- editing (boolean)
- commentId (Either the interpretation id, or the comment id of the interpretation comment.

Two more API functions are added, before dispatching the action:
- updateInterpretation
- updateInterpretationComment

Editing intepretations works fine, while editing interpretationComments updates data to the API incorrectly (The functions calls JSON.stringify on a text string).

Currently, d2.Api.getApi()**.update()**  within updateInterpretationComment updates the endpoint with additional quotations through JSON.stringify. This means that the payload will consist of quotations which the Regex in d2 API filters incorrectly.

Initially i tried to add helper functions to filter the string, however this is not the best solution.
I spoke with Birk regarding changing the d2 API call, but this will require a breaking API change as we thought the best change would be to add the option parameter for consistency.
(adding if statement to check if input is a string: https://github.com/dhis2/d2/tree/feature/validate-type-update-function). 

Postman updates the endpoint correctly when specified that the payload data is content-type: text-plain. But .update() does not have options available as parameters.


joakimia:
PR for performing the PUT request with the text/plain option: https://github.com/dhis2/d2/pull/165

I've also added an if statement to verify that the edit contains character, as the PUT request will otherwise prompt an error when the body of the payload is empty.

It also forces the user to delete the edit instead of leaving an empty field (should the attempt to delete the whole text and post the update).

src/Item/PluginItem/Interpretations/Interpretation.js
```diff
@@ -127,6 +139,45 @@ class Interpretation extends Component {
...
+    };
+
+    // new: joakim (find better name) -> Render InputField or existing text, if ID is present in Array.
+    renderCommentOrEditField = item => {
+        return this.state.editing.includes(item.id) ? (
+            <InputField
+                editing
+                placeholder={item.text}
+                commentId={item.id}
+                postText={'Post'}
```
edoardo: Remember i18n

package.json
```diff
@@ -45,6 +45,7 @@
     "jsdoc": "^3.5.5",
     "lint-staged": "^7.0.0",
     "material-ui": "^0.20.0",
+    "moment": "^2.22.1",
```
edoardo: is this needed?

src/Item/PluginItem/Interpretations/Interpretation.js
```diff
@@ -155,7 +205,7 @@ class Interpretation extends Component {
             ? i18n.t('You like this')
             : i18n.t('Like');
 
-        const canDeleteInterpretation = () =>
+        const canModifyInterpretation = () =>
             this.userIsOwner(this.props.interpretation.user.id) ||
             this.props.interpretation.access.delete ||
             this.props.userIsSuperuser;
```
edoardo: Are the conditions the same for both edit and delete?

src/api/interpretations.js
```diff
@@ -37,10 +37,11 @@ export const postInterpretationSharing = data => {
         .catch(onError);
 };
 
+// New: joakim
 export const updateInterpretation = data => {
     const url = `/interpretations/${data.id}`;
     return getInstance()
-        .then(d2 => d2.Api.getApi().update(url, data.text))
+        .then(d2 => d2.Api.getApi().patch(url, { text: data.text }))
```
edoardo: I guess with the fix in d2 this should go back to update?

edoardo (changes requested):
Remove the `new: Joakim` comments 😉 

Also, I think @jenniferarnesen should have a look at this as she worked with the original interpretations component.

src/Item/PluginItem/Interpretations/Interpretation.js
joakimia: Went through the files and added i18n.t 
Asked @jenniferarnesen to take a look as well

package.json
joakimia: Removed this dependency, it is not needed.

src/Item/PluginItem/Interpretations/Interpretation.js
joakimia: I will double check with Scott and @jenniferarnesen , Maybe the owner should have the possibility of both editing and deleting, and super user should only be able to delete.

I'll separate it for now 👍

src/api/interpretations.js
joakimia: Yes i agree, changed it back to .update(),

src/Item/PluginItem/Interpretations/Interpretation.js
```diff
@@ -127,6 +139,42 @@ class Interpretation extends Component {
...
+        return this.state.editing.includes(item.id) ? (
+            <InputField
+                editing
+                placeholder={i18n.t(item.text)}
+                commentId={item.id}
+                postText={i18n.t('Post')}
+                onEdit={this.editComment}
+            />
+        ) : (
+            <p style={style.text}>{i18n.t(item.text)}</p>
```
jenniferarnesen: @joakimia item.text is the actual comment, so there is no translation for it. So remove the i18n.t part.

src/Item/PluginItem/Interpretations/Interpretation.js
```diff
@@ -118,7 +130,7 @@ class Interpretation extends Component {
     };
 
     showCommentField = () => {
```
jenniferarnesen: The use of this has changed, so it would be good to rename it accordingly (e.g., toggleCommentField)

jenniferarnesen (approved):

PR #197, opened 2018-06-20
Feature/message-item-style Improve styling of messages item [DHIS2-2824]

jenniferarnesen:
Update styling of the dashboard messages item according to the sketch and comments in the Jira issue
* Remove 'All' and 'unread' filters
* Show unread messages first
* Include a link at the bottom of the scrolling to redirect to the messaging app
* Load up to 200 messages


janhenrikoverland (approved):

jenniferarnesen:
The reason the fn redirectToMsg is inside the render is because there is no other way to know what the message type and id for each message conversation are.

The alternative is to turn the li into a separate component, then I can use component props to get those values.

PR #201, opened 2018-06-26
fix/no-access - Clearly separate between new and edit mode [DHIS2-4091]

jenniferarnesen:
Fix bug where user could not create a new dashboard, because the access check was based on the access level for the dashboard being viewed when the new button was clicked. The solution was to clearly separate new and edit mode when rendering the dashboard.

The fix involved a bit of refactoring to have similar components for View, Edit, and New states.


src/Dashboard/Dashboard.js
```diff
@@ -23,18 +22,19 @@ class Dashboard extends Component {
...
-                <DashboardsBar />
-                <DashboardVerticalOffset />
-                <ViewDashboardContent />
-            </Fragment>
-        ) : (
-            <Fragment>
-                <EditDashboard mode={this.props.mode} />
-            </Fragment>
-        );
+        return <Fragment>{this.getDashboard()}</Fragment>;
```
janhenrikoverland: This Fragment is not necessary

janhenrikoverland: And now that render() is purely a function call it is probably over-extracted :-)

src/Dashboard/Dashboard.js
```diff
@@ -23,18 +22,19 @@ class Dashboard extends Component {
         this.setDashboard();
     }
 
+    getDashboard = () => {
+        switch (this.props.mode) {
+            case 'edit':
```
janhenrikoverland: We have 'edit', 'new', 'view' a couple of places in the app now. Could be centralized constants.

src/Dashboard/EditDashboard.js
```diff
@@ -39,12 +30,31 @@ class EditDashboard extends Component {
...
+                <NoContentMessage text={i18n.t('No access')} />
+            );
+        };
+
+        const contentNotReady =
+            !this.props.dashboardsLoaded || this.props.selectedId === null;
+
+        return (
+            <div className="dashboard-wrapper">
+                {contentNotReady ? null : <Content />}
```
janhenrikoverland: Just to be nitpicky the Content component could be created only if it's needed. Like:

{contentNotReady ? null : this.getContent(this.props.updateAccess)}

src/Dashboard/NewDashboard.js
```diff
@@ -0,0 +1,29 @@
...
+import { connect } from 'react-redux';
+import DashboardVerticalOffset from './DashboardVerticalOffset';
+import EditBar from '../ControlBarContainer/EditBar';
+import DashboardContent from './DashboardContent';
+import { acSetEditNewDashboard } from '../actions/editDashboard';
+
+class NewDashboard extends Component {
+    constructor(props) {
+        super(props);
+        this.props.setNewDashboard();
```
janhenrikoverland: Would expect this to happen in componentDidMount

src/Dashboard/ViewDashboard.js
```diff
@@ -0,0 +1,57 @@
...
+export const ViewDashboard = ({ id, dashboardsIsEmpty, dashboardsLoaded }) => {
+    const hasDashboardContent = id && !dashboardsIsEmpty;
+    const contentNotReady = !dashboardsLoaded || id === null;
+
+    return (
+        <Fragment>
+            <DashboardsBar />
+            <DashboardVerticalOffset />
+            <div className="dashboard-wrapper">
+                {contentNotReady ? null : (
```
janhenrikoverland, line 36: Ref preious comment - here you actually do this in an even better way.

janhenrikoverland (approved):

PR #203, opened 2018-07-02
Cancel request if type is already active (DHIS2-4186)

janhenrikoverland:
Check new active type against `current active type  ||  original visualization type`.

This works because current active type is `undefined` before first switching of type, but never after that, even when you switch back to original type.


jenniferarnesen (approved):

jenniferarnesen:
Please append a jira issue number to the commit msg when merging

PR #205, opened 2018-07-05
Toggle interpretation/comment edit button text (DHIS2-4208)

janhenrikoverland:
- Checks if the current interpretation/comment's id equals state.editId
- Converts editButton and deleteButton into stateless function components with props

### DIFF
--- a/src/Item/PluginItem/Interpretations/Interpretation.js
+++ b/src/Item/PluginItem/Interpretations/Interpretation.js
@@ -72,7 +72,7 @@ const style = {
     },
 };
 
-const deleteButton = action => {
+const DeleteButton = action => {
     const iconStyle = Object.assign({}, style.icon, { fill: colors.red });
 
     return (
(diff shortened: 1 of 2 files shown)


src/Item/PluginItem/Interpretations/Interpretation.js
```diff
@@ -270,10 +277,14 @@ class Interpretation extends Component {
...
-                    editButton(() => this.toggleEdit(comment.id))}
+                {this.userIsOwner(comment.user.id) && (
+                    <EditButton
+                        action={() => this.toggleEdit(comment.id)}
+                        text={this.getEditText(comment.id)}
+                    />
+                )}
                 {this.hasDeleteAccess(comment.user.id) &&
-                    deleteButton(() => this.deleteComment(comment.id))}
+                    DeleteButton(() => this.deleteComment(comment.id))}
```
edoardo: shouldn't this be `DeleteButton({action: () => this.deleteComment(comment.id)})`

edoardo (changes requested):

src/Item/PluginItem/Interpretations/Interpretation.js
janhenrikoverland: It should actually be <DeleteButton action={........................ Forgot to update this instance.

edoardo (approved):

PR #224, opened 2018-12-14
Prepare for new chart plugin component

janhenrikoverland:
- Extract visualization part of the visualization item, keep header and footer in the parent component
- If CHART, use the new chart component, otherwise use the default visualization item
- Centralize and reuse the "pluginIsAvailable" function


edoardo (approved):
LGTM (after fixing the conflicts) 😉

PR #225, opened 2018-12-17
chore/named imports

janhenrikoverland:
- Remove the fromX pattern (e.g. fromReducers.fromDashboard.xxx) and import the actual variables instead
- Remove the actionTypes objects and import/export all actions instead
- Rename all default state variables to "DEFAULT_STATE_xxx"
- Rename all root selectors from e.g. "sGetFromState" to e.g. "sGetDashboardsRoot"
- Remove an unused action creator
- Rename a misspelled selector

### DIFF
--- a/src/reducers/index.js
+++ b/src/reducers/index.js
@@ -1,38 +1,27 @@
 import { combineReducers } from 'redux';
 import arraySort from 'd2-utilizr/lib/arraySort';
-import dashboards, * as fromDashboards from './dashboards';
-import selected, * as fromSelected from './selected';
-import dashboardsFilter, * as fromDashboardsFilter from './dashboardsFilter';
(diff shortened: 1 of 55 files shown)


src/reducers/index.js
```diff
@@ -49,42 +38,26 @@ export default combineReducers({
...
 
-// filter dashboards by name
+// Filter dashboards by name
 export const sFilterDashboardsByName = (dashboards, filter) =>
     dashboards.filter(d =>
         d.displayName.toLowerCase().includes(filter.toLowerCase())
     );
 
-// filter dashboards by owner
+// Filter dashboards by owner, TODO FIXME
```
jenniferarnesen, line 57: Are these TODO's something that can be fixed now, in a separate PR?

jenniferarnesen (approved):

PR #227, opened 2018-12-17
Remove unused widgets

janhenrikoverland:
- Remove D2ContentEditable and D2DropDown


src/widgets/D2ContentEditable.js
```diff
@@ -1,62 +0,0 @@
-import React from 'react';
-import PropTypes from 'prop-types';
-import ContentEditable from 'react-contenteditable';
```
jenniferarnesen, line 3: Make sure to remove this dependency as well.

janhenrikoverland: Thanks

jenniferarnesen (approved):

PR #243, opened 2019-02-06
chore: remove unnecessary development proxy

amcgee:
The proxy setup in package.json appears to be unnecessary, and breaks if the core is running anywhere other than `localhost:8080`.  The proxy was only used for `dhis-web-core-resources/fonts/roboto.css`, and other `core-resources` references are inserted as vendor scripts with the fully qualified DHIS2 baseUrl, so I've removed the proxy and used baseUrl instead.

I also added conditional support for `REACT_APP_DHIS2_BASE_URL` if no dhis_config is specified, before falling back to `localhost:8080`


amcgee:
Ping @jenniferarnesen @janhenrikoverland should be a quick review - any reason not to merge this?  It's a small change but fixes a bunch of errors logged when not pointing at localhost:8080

PR #246, opened 2019-02-13
Show user avatar in the headerbar

HendrikThePendric:
`@dhis2/ui` version `1.0.0-beta.9` contains a fix that will show the user avatar (profile picture) in the HeaderBar if one is found. I've updated the package and imports so that is now used here.


joakimia (commented):
Tested the branch, picture of user avatar pic are loaded correctly with the new headerBar:
image

The new reset.css file also allows rich text:
image

No errors provoked.
This can go in unless @jenniferarnesen have additional comments.

joakimia (approved):

PR #249, opened 2019-03-06
add RichText Parser to VisualizationItem description [DHIS2-3422]

joakimia:
Related to ticket [3422](https://jira.dhis2.org/browse/DHIS2-3422).

Rich text are being rendered in the dashboard, and in the apps. But the description field are missing.
I've also added `pre-line` to the Parser in order to preserve the original format with carriage return.


edoardo (approved):
LGTM

PR #260, opened 2019-03-08
Fix dashboard search bug

janhenrikoverland:
When list of dashboards is empty, don't try to redirect to the first one.


edoardo (approved):
LGTM

PR #263, opened 2019-03-11
chore: update to latest cra and no eject

jenniferarnesen:
Restore the latest create-react-app and do not eject

Changes include:
* updated to react 16.8, webpack 4, and other upgrades
* remove jest config (package.json), and webpack configs, as they are now embedded in react-scripts configurations
* add a script loader to src/index.js to syncronously load scripts previously added in webpack htmlWebpackPlugin.vendorScripts
* read manifest file using d2.getManifest, and set config at runtime
* require env vars REACT_APP_DHIS2_BASE_URL and REACT_APP_DHIS2_AUTHORIZATION at runtime for dev mode
* App.js is now a redux-connected component, to work with react-redux >=6.0
* Other tweaks to src files and tests to work with dependency upgrades
* for the build, it was necessary to add `PUBLIC_URL=.` otherwise, the urls were absolute based on the root of the app, resulting in:`http://localhost:8080/static/js/2.ad42edc8.chunk.js` instead of ` http://localhost:8080/dhis-web-dashboard/static/js/2.ad42edc8.chunk.js`

Remaining issues:
* ~~map plugin does not work (complains of `React not defined`) - issue reported to Bjørn and Austin~~
* ~~dv plugin does not work - I think this is due to the reference to `analytics-shared` in the plugin~~
* ~~the font woff files are not downloading (getting a 302 and no content) - do we need the proxy after all?~~
* ~~consider adding a prettier pre-commit hook as described in CRA documentation~~ (will add cli-style in a separate PR)
* ~~Is this still necessary in the index.html file? `<script>jQuery.migrateMute = true;</script>`~~


varl:
>the font woff files are not downloading (getting a 302 and no content) - do we need the proxy after all?

Bundle the fonts:

```
yarn add typeface-roboto
```

then in your entrypoint for the App:

```
import 'typeface-roboto'
```

>consider adding a prettier pre-commit hook as described in CRA documentation

A good opportunity to set up: https://github.com/dhis2/cli-style

>Is this still necessary in the index.html file? <script>jQuery.migrateMute = true;</script>

Shouldn't be, I've removed it when going over https://github.com/dhis2/notes/issues/26 without any problems so far.

src/index.js
```diff
@@ -31,21 +29,69 @@ const configI18n = userSettings => {
     i18n.changeLanguage(uiLocale);
 };
 
-const init = () => {
+const injectScriptsIntoBody = scriptPrefix => {
```
varl: I suggest bundling these artifacts. The core-resources will be deprecated in 2.32 and removed in 2.33.

Do we need jquery still?

varl (commented):
This looks really good!

I have some suggestions to bring it closer to https://github.com/dhis2/notes/issues/26 otherwise I like the direction.

jenniferarnesen:
> A good opportunity to set up: https://github.com/dhis2/cli-style

Will discuss the changes with the team and do this in a separate PR since it will change pretty much every line of every src file.

varl:
> 
> 
> > A good opportunity to set up: https://github.com/dhis2/cli-style
> 
> Will discuss the changes with the team and do this in a separate PR since it will change pretty much every line of every file.

Only if you run `--all`, by default it just applies the code style to staged files.

It does replace your Prettier configuration however, so if you have an editor which reads that and then reformat an entire file with the editor, it will reformat to the new code style at that point.

src/index.js
jenniferarnesen: The plugins (`eventreport`, `eventchart`, `reporttable`, and `maps`) are not yet available as published libraries. So that will have to come later.

As far as jQuery, for now, we could get those from another CDN.

I'm not sure we need to include the babel-polyfill actually. Doesn't CRA bundle all the needed polyfills in the build?

jenniferarnesen: > I suggest bundling these artifacts. The core-resources will be deprecated in 2.32 and removed in 2.33.

The plugins (eventreport, eventchart, reporttable, and maps) are not yet available as published libraries. So that will have to come later.

As far as jQuery, for now, we could get those from another CDN. jQuery is unfortunately needed by the older plugins.

I'm not sure we need to include the babel-polyfill actually. Doesn't CRA bundle all the needed polyfills in the build?

varl: >The plugins (eventreport, eventchart, reporttable, and maps) are not yet available as published libraries. So that will have to come later.

That's fine, they look like they are linked to the app, not to core-resource. Those are the ones we need to get rid of first and foremost. 

CRA polyfill should do nicely unless there is code which uses something special.

Relying on an external CDN is not a good idea. Some of the DHIS2 instances are on isolated networks which may or may not have external internet access. Better to bundle it with the app or the plugin. If it's required it _is_ a dependency.

jenniferarnesen: Yeah

> Relying on an external CDN is not a good idea

Yeah, I realized that right after I wrote it - thought I changed my comment, but apparently not in time. I'll look into bundling jQuery and see what is needed from babel-polyfill (Bjørn says Maps needs it)

amcgee: > CRA polyfill should do nicely unless there is code which uses something special.

I think we need to include `import 'react-app-polyfill/ie11';` from [CRA](https://github.com/facebook/create-react-app/tree/master/packages/react-app-polyfill) to support ie11?

> see what is needed from babel-polyfill (Bjørn says Maps needs it)

(note: maps doesn't use CRA currently, see https://github.com/dhis2/maps-app/pull/89 which I worked on a while ago)

README.md
```diff
@@ -10,8 +10,12 @@ It was bootstrapped with [Create React App](https://github.com/facebookincubator
...
 ### Prerequisites
 
-To use the dashboards-app in development mode, it is necessary to have a running DHIS2 instance, and be logged in to it. This is because the app requests resources, like react and react-dom, from the DHIS2 core resources. Most developers use their local DHIS2 instance. If you haven't configured your DHIS2_HOME env variable and set up a config.json, then the app will default to using localhost:8080 with the admin user (see
-[config/webpack.config.dev.js](config/webpack.config.dev.js#L35)).
+To use the dashboards-app in development mode, it is necessary to have a running DHIS2 instance, and be logged in to it. This is because the app requests resources, like maps plugin, from the DHIS2 core resources. Most developers use their local DHIS2 instance.
+
+The following env vars need to be set when running in dev mode:
+
+-   REACT_APP_DHIS2_BASE_URL (e.g., http://localhost:8080)
+-   REACT_APP_DHIS2_AUTHORIZATION (base64 encoded username and password)
```
amcgee, line 18: I don't like requiring this to be an environment variable, isn't it better to just rely on cookies and manual login?  Otherwise you probably end up only testing with a single user.

jenniferarnesen: The analytics plugins need the password in order to work (in dev mode), unfortunately

src/index.js
varl: > I think we need to include `import 'react-app-polyfill/ie11';` from [CRA](https://github.com/facebook/create-react-app/tree/master/packages/react-app-polyfill) to support ie11?

Oh, shoot. Yeah they changed that in CRA2.0. You're correct.

README.md
amcgee: @jenniferarnesen why is that?

jenniferarnesen: I'm not sure - Jan would know. But my testing confirms that the plugins don't work without a password.

jenniferarnesen:
@amcgee @varl 

Regarding the core-resource-app dependencies (`babel-polyfill`, `jquery`, `jquery-migrate`), I tried importing `jquery` and `jquery-migrate` as es modules but that didn't work out of the box (and I was tired). Would greatly appreciate if someone else would give this a go. Maybe it's premature to remove core-resources from dashboards, and we should instead shoot for 2.33.

varl:
> 
> 
> @amcgee @varl
> 
> Regarding the core-resource-app dependencies (`babel-polyfill`, `jquery`, `jquery-migrate`), I tried importing `jquery` and `jquery-migrate` as es modules but that didn't work out of the box (and I was tired). Would greatly appreciate if someone else would give this a go. Maybe it's premature to remove core-resources from dashboards, and we should instead shoot for 2.33.

Is anything in Dashboards using `jquery` and `jquery-migrate`?

I cannot find anything. If Dashboards just loads jQuery to give all the plugins access to it on the `window` object I think those plugins should have the dependency to jQuery and only include the parts they need from it. A plugin should be able to just _plug in_.

I'm getting a bunch of CORS error when trying to load plugins locally though, so I'm having problems with setting up the local environment for Dashboards.

package.json
```diff
@@ -113,66 +67,32 @@
...
     "devDependencies": {
-        "@dhis2/d2-i18n-generate": "^v1.1.0",
-        "create-react-class": "^15.6.3",
-        "d2-i18n-extract": "^1.x",
-        "enzyme-adapter-react-16": "^1.7.1",
-        "husky": "^0.15.0-rc.12",
-        "jest-enzyme": "^7.0.1",
-        "react-test-renderer": "^16.7.0",
-        "webpack-bundle-analyzer": "^3.0.3"
+        "@dhis2/d2-i18n-extract": "^1.0.7",
```
amcgee, line 90: This `d2-i18n-extract` replaced `d2-i18n-generate` which doesn't seem like it should work - both are required and  referenced in scripts...

public/index.html
```diff
@@ -1,29 +1,49 @@
...
+        <script src="../dhis-web-core-resource/babel-polyfill/6.20.0/dist/polyfill.min.js"></script>
+        <script src="../dhis-web-core-resource/jquery/3.2.1/dist/jquery.min.js"></script>
+        <script src="../dhis-web-core-resource/jquery-migrate/3.0.1/dist/jquery-migrate.min.js"></script>
+        <script src="../dhis-web-pivot/reporttable.js"></script>
+        <script src="../dhis-web-maps/map.js"></script>
+        <script src="../dhis-web-event-reports/eventreport.js"></script>
+        <script src="../dhis-web-event-visualizer/eventchart.js"></script>
+        <% } %> <% if (process.env.NODE_ENV !== 'production') { %>
+        <script src="%REACT_APP_DHIS2_BASE_URL%/dhis-web-core-resource/babel-polyfill/6.20.0/dist/polyfill.min.js"></script>
+        <script src="%REACT_APP_DHIS2_BASE_URL%/dhis-web-core-resource/jquery/3.2.1/dist/jquery.min.js"></script>
```
amcgee: As an interim solution shouldn't we be able to include the static `jquery` files and reference them locally here?

That would still expose them in the global namespace and doesn't require bundling into each plugin (though we should eventually figure out a better solution for that) while removing the core-resource dependency.

amcgee (changes requested):

public/index.html
varl: Yes.

amcgee: @varl @jenniferarnesen I just pushed a commit with this change.

varl: This brings me back to 2008. <3

public/index.html
```diff
@@ -15,17 +15,22 @@
...
         <script src="../dhis-web-pivot/reporttable.js"></script>
         <script src="../dhis-web-maps/map.js"></script>
         <script src="../dhis-web-event-reports/eventreport.js"></script>
         <script src="../dhis-web-event-visualizer/eventchart.js"></script>
         <% } %> <% if (process.env.NODE_ENV !== 'production') { %>
-        <script src="%REACT_APP_DHIS2_BASE_URL%/dhis-web-core-resource/babel-polyfill/6.20.0/dist/polyfill.min.js"></script>
-        <script src="%REACT_APP_DHIS2_BASE_URL%/dhis-web-core-resource/jquery/3.2.1/dist/jquery.min.js"></script>
-        <script src="%REACT_APP_DHIS2_BASE_URL%/dhis-web-core-resource/jquery-migrate/3.0.1/dist/jquery-migrate.min.js"></script>
+        <!-- TODO: Remove these global script dependencies or bundle them with their respective plugins -->
+        <script src="%PUBLIC_URL%/vendor/babel-polyfill-6.26.0.js"></script>
```
jenniferarnesen: Doesn't this need to be %REACT_APP_DHIS2_BASE_URL%  ?

jenniferarnesen:
@amcgee This works great for me, both in dev and prod mode.  I say we go ahead with this version, and do more improvements later (code splitting, removing old MUI, adding Cypress) etc in new PRs.

There is still the issue of all the failing Netlify checks. Do you have time to see what the issue is on those?

@varl Do you still have problems getting this branch to work?

public/index.html
jenniferarnesen: You know would be super cool is to remove the dependency on jQuery (not now, but down the road). Not sure what that would entail.

package.json
jenniferarnesen: Looks like at least one of the d2-ui packages installed it. I've re-added it as a dev dependency.

varl:
@jenniferarnesen I had to add `d2-i18n` as a dependency, but I'm still getting the following error locally: 

```
Module not found: Can't resolve 'moment/locale/' in '/home/varl/dev/dhis2/apps/dashboards-app/src/locales'
```

amcgee:
@jenniferarnesen @varl

1) It looks like we have two `d2-i18n-generate` packages, one at [@dhis2/d2-i18n-generate](npmjs.org/@dhis2/d2-i18n-generate) and an older one at [d2-i18n-generate](npmjs.org/d2-i18n-generate).  I've updated the dependency here, we should deprecate the un-namespaced package.

2) @varl I think the error you were seeing was because of a missing `moment` dependency, I've added it here.  For some reason I wasn't getting the error locally (probably because `moment` resolved somewhere else on my machine).

3) @jenniferarnesen looking into the failed builds now.

public/index.html
amcgee: @jenniferarnesen FWIW we use `%PUBLIC_URL%` to reference files relative to the app root, rather than the DHIS2 core root.  We don't want to assume that we're installed at a particular sub-url on the DHIS2 server (though currently we are and are using those URLs to reference the plugins)

amcgee (approved):

amcgee:
@jenniferarnesen @varl this should be good to go now, I fixed the netlify build issue (it was referencing the old `scripts/build.js` which I now removed).

The netlify deployment itself does not work because of the `../` hack for plugin URLs in production as well as a pesky CORS stickiness, but I don't know if people are actually using netlify here.

We'll want to fix that, but I think it's better to look at plugin loading holistically rather than trying to jerry-rig something here, do you both agree?

public/index.html
jenniferarnesen: Yepp.

jenniferarnesen:
`moment` was not a dependency previoulsy.

I'm not convinced that `moment` should be a dependency. It is a dependency of `d2-i18n-generate`, which writes the `src/locales` files. Well, I guess you could argue that it should be a dependency since the result of d2-i18n-generate is files in `src` with import from moment... Not sure.

In the second case, maybe d2-i18n-generate should indicate `moment` as a peerDependency.

jenniferarnesen:
> a pesky CORS stickiness, but I don't know if people are actually using netlify here.

We were starting to use Netlify for checking other browsers, but it definitely hadn't become habit quite yet. So fine with me to go this way for now. 

It will definitely be a goal to get rid of the '../' hack. One solution is to publish the plugins (as dv plugin is "published"). I suppose we could turn the hardcoded '../' in the template also into a  `REACT_APP` env var, but not sure if that would solve anything for netlify. Also, its one more env var that is needed.

amcgee:
> `moment` was not a dependency previoulsy.
> 
> I'm not convinced that `moment` should be a dependency. It is a dependency of `d2-i18n-generate`, which writes the `src/locales` files. Well, I guess you could argue that it should be a dependency since the result of d2-i18n-generate is files in `src` with import from moment... Not sure.
> 
> In the second case, maybe d2-i18n-generate should indicate `moment` as a peerDependency.

Yeah, I don't like that `d2-i18n-generate` produces source files in the consumer, that seems wrong.  It should probably declare the peer dependency, but that gets a little weird because `d2-i18n-generate` is a devDep and `moment` would need to be a runtime dep.  But it's probably better than not declaring anything...



> We were starting to use Netlify for checking other browsers, but it definitely hadn't become habit quite yet. So fine with me to go this way for now.
> 
> It will definitely be a goal to get rid of the '../' hack. One solution is to publish the plugins (as dv plugin is "published"). I suppose we could turn the hardcoded '../' in the template also into a `REACT_APP` env var, but not sure if that would solve anything for netlify. Also, its one more env var that is needed.

I would love to get Netlify working!  Using an env var might be a hack since we would bake the '..' into the built production artifact.  I could try creating a proxy in netlify, though it will have to bake in the authentication credentials to retrieve the plugin code

amcgee:
@jenniferarnesen I ran with your env var idea and made things a bit cleaner, see my last commit.  Basically I continued to use `REACT_APP_DHIS2_BASE_URL` in production as well as development, and set that variable in git-committed `.env` and `.env.production` files appropriately.  We can then override the value in a production build (see `package.json::build-netlify`) and it will also be picked up by the manifest generation (`package.json::manifest`[1]).

A few caveats:
1. The env var syntax in package.json scripts is posix-only, but we were already incompatible with building on windows so I think that's OK for now.
2. I'm not setting `REACT_APP_DHIS2_AUTHORIZATION` in production or in netlify, we might need to add that
3. We're still using the `..` hack in the published production artifact, but I don't think we have a choice at this point.

Netlify deployment [is working](https://deploy-preview-263--dhis2-dashboards.netlify.com/#/).  The AUTHORIZATION env var shouldn't be required, will need to investigate that.

CC @varl

varl (approved):

amcgee:
@jenniferarnesen want to do the honors of merging this?  Unless @janhenrikoverland or @edoardo have comments to add?

amcgee:
I'll resolve the merge conflicts here and test one more time so we can get this in.  There's an issue with pivot tables not supporting remote Core servers (a la Netlify) but it works when installed in a WAR file (will confirm that again).  @janhenrikoverland want to give a thumbs-up if you think this is OK to merge?

amcgee:
Updates from `master` have been merged, demo is available on [netlify](https://deploy-preview-263--dhis2-dashboards.netlify.com).  Note that since we reverted to the old Charts plugin, charts also render blank for the same reason as pivots (confirming that they work in a WAR file now)

janhenrikoverland (approved):

amcgee:
The chart plugin doesn't work even in a war file, though pivots table does.  I'll have to investigate the discrepancy before merging this

jenniferarnesen:
> The chart plugin doesn't work even in a war file, though pivots table does. I'll have to investigate the discrepancy before merging this

@amcgee Any update on this?

amcgee:
@jenniferarnesen this is I believe related to the change to use the old visualizer plugin, I didn't get further into debugging.  @neeilya was maybe going to take a look since he made the change that was merged into this branch and caused the issue.  @neeilya have you had any thoughts?

jenniferarnesen:
> 2\. I'm not setting `REACT_APP_DHIS2_AUTHORIZATION` in production or in netlify, we might need to add that

I confirm that the AUTHORIZATION is not needed in production. Here in `src/index.js`

```
config.headers = isProd ? null : { Authorization: authorization };
```

PR #264, opened 2019-03-12
Fix: visualization item drop down crash

joakimia:
This PR includes validating that the description exist before passing it to the rich text parser.
The crash are caused when opening a visualization without an existing description.


neeilya (approved):
Thanks for the quick fix!

PR #267, opened 2019-03-13
[WIP] Fix switching visualization type for dashboard items

neeilya:
Fixes [DHIS2-5997](https://jira.dhis2.org/browse/DHIS2-5997).

This PR contains fix for switching visualization type for dashboard items.


neeilya:
We decided to change the approach for fetching analytical objects (AO).

The plan is to fetch data on app level and pass complete AOs down to plugins for rendering. This approach will eliminate visualization type switching problem

PR #269, opened 2019-03-19
Fixing visualization type switch for dashboard items

neeilya:
Fixes [DHIS2-5997](https://jira.dhis2.org/browse/DHIS2-5997).

This PR fixes visualization type change feature by using old plugin component `DefaultPlugin.js`.

The reason for switching to the old plugin is that we will fetch data on app level and pass complete analytic objects (AO) to plugins for rendering. This will reduce number of network requests and allow to correctly use `ChartPlugin` component.


edoardo (approved):
LGTM

PR #271, opened 2019-03-21
feat: fetch items chart data on app level

neeilya:
Relates to [DHIS2-6320](https://jira.dhis2.org/browse/DHIS2-6320).

Changes include:
- New approach for rendering chart items: fetching full analytical object (AO) on app level and passing it down to plugins to make filtering possible and avoid redundant network requests.
- Using progressive loader for `Visualization/Item` component instead of its content component, so that fetching AO happens only when component is mounted.
- Mounting chart to div inside `DefaultPlugin` component for correct further unmount.
- Reenabling new `ChartPlugin` component for CHART visualization type.
- Pass dynamic height via styles prop to plugins instead of setting height of div containing the plugin.

### DIFF
--- a/src/components/Item/VisualizationItem/DefaultPlugin.js
+++ b/src/components/Item/VisualizationItem/DefaultPlugin.js
@@ -3,8 +3,9 @@ import PropTypes from 'prop-types';
 import i18n from 'd2-i18n';
 
 import * as pluginManager from './plugin';
-import { getBaseUrl, orObject } from '../../../modules/util';
+import { getBaseUrl, orObject, withoutId } from '../../../modules/util';
 import { getGridItemDomId } from '../../ItemGrid/gridUtil';
+import { itemTypeMap } from '../../../modules/itemTypes';
 
 const pluginCredentials = d2 => {
     return {
@@ -15,6 +16,12 @@ const pluginCredentials = d2 => {
 
 class DefaultPlugin extends Component {
     pluginCredentials = null;
+
+    constructor(props, context) {
+        super(props);
+
+        this.d2 = context.d2;
+    }
 
     shouldPluginReload = prevProps => {
         // TODO - fix this hack, to handle bug with multiple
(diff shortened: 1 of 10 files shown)


src/components/Item/VisualizationItem/DefaultPlugin.js
```diff
@@ -95,6 +97,11 @@ class DefaultPlugin extends Component {
     getActiveType = () =>
         this.props.visualization.activeType || this.props.item.type;
 
+    getChartConfig = () => {
```
edoardo: Why `Chart` in the name? this will never be a chart AO...

src/components/Item/VisualizationItem/DefaultPlugin.js
```diff
@@ -133,11 +142,15 @@ DefaultPlugin.contextTypes = {
 };
 
 DefaultPlugin.propTypes = {
+    editMode: PropTypes.bool,
```
edoardo: Is this used in `DefaultPlugin`?

src/components/Item/VisualizationItem/Item.js
```diff
@@ -135,6 +152,22 @@ export class Item extends Component {
             : null;
     };
 
+    getPluginComponent = () =>
+        this.getActiveType() === CHART ? (
+            <div>
```
edoardo: Do we need these extra `<div>` elements?

neeilya: Thanks for reporting, forgot to remove these after checking some things.

src/components/Item/VisualizationItem/DefaultPlugin.js
neeilya: Looks like we [used it before](https://github.com/dhis2/dashboards-app/blob/master/src/components/Item/VisualizationItem/DefaultPlugin.js#L83) and I updated props, but then removed the line where it was used.

Will update props again now.

edoardo (approved):

neeilya:
Will merge as soon as `d2-charts-api` library version `31.0.14` with bugfix is published (currently publishing fails because of the old api key, resolving now).

PR #274, opened 2019-04-02
v31 - fix: pin react version to prevent broken sharing dialog

amcgee:
This fixes [DHIS2-6506](https://jira.dhis2.org/browse/DHIS2-6506) by upgrading to a modern sharing dialog.  Previously this used `d2-ui-sharing` which is more than a year old and doesn't play well with our current version of react.

Some caveats about this fix:
1 - the dialog looks a bit weird (no minimum width) because we don't have the DHIS2 Mui theme in Dashboards.  It's disruptive to add it, so since the dialog is functional and not SUPER ugly I'm advocating for leaving it as-is (CC @joeDHIS)
2 - there is a horrendous hodgepodge of `d2`, `d2-ui`, and `@dhis2/d2-ui-*` components in here which means there's a lot of duplication and bundle bloat and probably hidden bugs.  I'm not touching any of that with this change, just adding the dep on `@dhis2/d2-ui-sharing-dialog` which in turn depends on `@dhis2/d2-ui-core` and `d2@31` (dashboards itself uses `d2-ui` and `d2@30`).  Both of those indirect dependencies are already included in the `dashboards` bundle by the modern `@dhis2/d2-ui-interpretations` dependency, but we **really** need to think more carefully about how we manage the spiderweb in the future.  I think this change is safe to fix a major bug in 2.31, and it shouldn't bloat the bundle any more than it already is bloated, but it should be thoroughly tested.

The working dialog:

Screenshot 2019-04-02 16 20 56

For the record, here's a snapshot of the awful dependency mess (cc @varl):

```
├─ @dhis2/d2-i18n-extract@1.0.7
├─ @dhis2/d2-i18n-generate@1.1.0
├─ @dhis2/d2-i18n@1.0.4
├─ @dhis2/d2-ui-core@0.0.0-PLACEHOLDER
│  └─ d2@31.4.0
├─ @dhis2/d2-ui-interpretations@5.2.10
│  └─ d2@31.4.0
│  ├─ @dhis2/d2-ui-core@5.2.10
│  ├─ @dhis2/d2-ui-mentions-wrapper@5.2.10
│  ├─ @dhis2/d2-ui-rich-text@5.2.10
│  └─ @dhis2/d2-ui-sharing-dialog@5.2.10
├─ @dhis2/d2-ui-mentions-wrapper@1.0.3
├─ @dhis2/d2-ui-rich-text@1.0.0
└─ @dhis2/d2-ui-sharing-dialog@5.2.11
├─ d2-i18n-extract@1.0.5
├─ d2-i18n-generate@1.0.6
├─ d2-i18n@1.0.3
├─ d2-manifest@1.0.0
├─ d2-ui@29.0.31
├─ d2-utilizr@0.2.16
└─ d2@30.1.0
```


jenniferarnesen:
Works fine. Too bad about the css issues, but I agree that its better that it works.

jenniferarnesen (approved):

Philip-Larsen-Donnelly:
I approve of having this dialog in just about any form, rather than a white screen of death. I am just curious what changed since 2.31.2, where the (old) sharing dialog is working?

varl:
The d2@31 and d2@30 are concerning. I've seen some issues related to two instances of d2 being loaded at the same time. There is usually a `d2.setInstance` override and/or the strategy to pass d2 as a prop to get around that, but it is a fragile situation.

How are the d2-i18n packages being pulled in twice on the same level?

amcgee:
@Philip-Larsen-Donnelly great call querying about the root cause, I thought this was in 2.31.2 as well but it looks like you're correct.

The culprit of this issue turned out to be 240b506ccb9e2d6da5fa1826fdd89e15f975e2b1 which required us to pin down a specific version of `d2-i18n-generate`.  While troubleshooting that issue it looks like the entire `yarn.lock` file was deleted and re-generated, causing the included version of React to be upgraded from `16.3` to `16.8`, since both satisfied our declared dependence on `^16.1` (anything up to `17`, the next major "breaking" change).  This React update apparently was not innocuous, causing this issue with Sharing Dialog.

To address this in a least-disruptive way, I've updated this PR.  Now, instead of upgrading `@dhis2/d2-ui-sharing-dialog` I am pinning the React version to `~16.3` which will keep it below `16.4`.  I've also reverted the `yarn.lock` change from 240b506ccb9e2d6da5fa1826fdd89e15f975e2b1 so that the diff of all dependencies from `2.31.2` to `2.31.3` is only `d2-i18n-generate` and `react`/`react-dom` (plus a patch version upgrade to `upath` so that we can use `node@10` when building).  This should mean that we don't run into any sneaky issues resulting from dependency changes.  Note that any bugs in `2.31.2` that might have been handled by minor version updates in dependencies won't be addressed, but since we don't know of any of those I think that's fine.

Updated this PR with the new approach, cleared @jenniferarnesen's review and requesting re-review of the new change.  The dialog looks a lot better (same as in `2.31.2`) and functions well in my testing.

amcgee:
@varl I agree wholeheartedly that the multiple versions of `d2` are problematic.  I *think* what got us into trouble here was back-porting the rich text interpretations editor to v31 (it's in `2.31.2`), because that latest interpretations component relies on `d2@31` and `@dhis2/d2-ui-sharing-dialog` and `@dhis2/d2-ui-core`.  You can actually see the same, scrunched sharing dialog when clicking `manage sharing` in an interpretation.  The interpretations backport may also have introduced multiple `material-ui` versions for the first time, though I haven't confirmed that.

I've updated this PR to pin `react` to `~16.4` instead of `~16.3` since this doesn't break the (old) sharing dialog but does satisfy the peer dependencies of about half a dozen DHIS2 packages (mostly related to the above interpretations change).

amcgee:
@varl regarding the multiple `d2-i18n-*` packages - I don't know why the namespaced ones are there, dashboards only relies on `d2-i18n-generate` and `d2-i18n-extract` not `@dhis2/d2-i18n-generate` nor `@dhis2/d2-i18n-extract`.  The former is definitely what's being called by the prebuild scripts.  Might be an issue with `yarn list` and caching or something?  Unclear, but (new theory) it's possible they're hoisted from multiple dependencies which incorrectly include them as dependencies rather than dev dependencies.

jenniferarnesen (approved):

jenniferarnesen:
Looks good and works.

PR #285, opened 2019-04-16
fix: Optimise item grid component to avoid unnecessary re-renders

neeilya:
Fixes [DHIS2-6597](https://jira.dhis2.org/browse/DHIS2-6597).

#### Problem:
Chart items re-render in dashboards app even if props stay the same (triggers by editing text item).

#### Previous proposal:
Make props deep comparison in [Chart plugin](https://github.com/dhis2/data-visualizer-app/pull/254).

#### New proposal:
Use `shouldComponentUpdate` function in `ItemGrid` component to determine if it should re-render the grid.

Apply comparison checks to following props:
- `dashboardItems`
- `edit`
- `isLoading`

### DIFF
--- a/src/components/ItemGrid/ItemGrid.js
+++ b/src/components/ItemGrid/ItemGrid.js
@@ -69,7 +69,10 @@ export class ItemGrid extends Component {
 
     shouldComponentUpdate(nextProps, nextState) {
         return (
-            !isEqual(this.props.dashboardItems, nextProps.dashboardItems) ||
+            !isEqual(
+                this.props.dashboardItems.length,
+                nextProps.dashboardItems.length
+            ) ||
             !isEqual(this.props.edit, nextProps.edit) ||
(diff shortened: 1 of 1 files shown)


jenniferarnesen:
Another test case: Expand the interpretations section of the item. Without the fix, the chart is also re-rendering there.

src/components/ItemGrid/ItemGrid.js
```diff
@@ -66,6 +67,18 @@ export class ItemGrid extends Component {
         this.props.acRemoveDashboardItem(clickedId);
     };
 
+    shouldComponentUpdate(nextProps, nextState) {
+        return (
+            !isEqual(
```
jenniferarnesen, line 72: How likely is the scenario where the number of items hasn't changed but the items are not the same?

neeilya: With current setup (only possible to add/remove one item at a time) I'd say it is impossible.

PR #286, opened 2019-04-23
fix: item-selector fixes

jenniferarnesen:
Fixes for:
- multi-dragging
- scrollbar issues
- sorting issues


edoardo:
PR #287 requires d2-ui-analytics@0.0.17 so if we merge that one this will come automatically.
We can also merge this, but then it must be done first.

jenniferarnesen:
d2-ui-analytics will get upgraded in another PR

PR #298, opened 2019-04-29
fix: set focus properties so textfield maintains focus when popover opens - v32

jenniferarnesen:
Fixes [DHIS2-6751]

Set the following Material-ui modal properties:

* disableAutoFocus: maintain focus in filter text field when item select Popover opens
* disableRestoreFocus: remove focus from filter text field when Popover closes (to match pre-existing behaviour)


edoardo (approved):

jenniferarnesen:
Fixes [DHIS2-6751]

Set the following Material-ui modal properties:

* disableAutoFocus: maintain focus in filter text field when item select Popover opens
* disableRestoreFocus: remove focus from filter text field when Popover closes (to match pre-existing behaviour)

PR #303, opened 2019-05-02
fix: set filter badge top position based on offset height [v32] (#301)

jenniferarnesen:
Fixes [DHIS2-6631]

The top position of the filter badges should be based on the height of the dashboards bar, so that the dashboards bar doesn't hide the badge if its been configured to more than one row


jenniferarnesen:
This one is on hold because of icon alignment issues described here: https://github.com/dhis2/ui-core/issues/194

edoardo (approved):

PR #305, opened 2019-05-02
chore: replace old MUI Snackbar component with new MUI

jenniferarnesen:
As a quick test, open the dashboard, then throttle the network connection to Slow 3G, then switch to another dashboard. The snackbar should appear at the bottom of the viewport.


neeilya (approved):
Tested the code, works as expected.

PR #306, opened 2019-05-02
chore: replace old mui add button with html button

jenniferarnesen:
Getting rid of old material-ui (0.20)

This is the green + button in the upper left of the dashboard. You can look at it on different browsers at this url:  https://deploy-preview-306--dhis2-dashboards.netlify.com


cooper-joe (approved):

src/widgets/AddDashboardButton.js
```diff
@@ -0,0 +1,38 @@
+import React from 'react';
+import AddIcon from '@material-ui/icons/AddCircle';
+import { colors } from '../modules/colors';
+
+const AddDashboardButton = () => {
+    return (
```
amcgee: Minor style issue but wouldn't this be cleaner without the functional body?  I.E. `const AddDashboardButton = () => (<React.Fragment>...</React.Fragment>)`

amcgee (changes requested):
This apprears to work as a global `<style>` but doesn't actually use `styled-jsx` because CRA needs an additional babel plugin to support it.  I'll look into how that might be achieved, but it may be non-trivial.

src/widgets/AddDashboardButton.js
jenniferarnesen: Done

jenniferarnesen:
@amcgee 
So I've gone through some iterations to arrive at the current solution:
- 1) moved css to a separate file and used css modules. This led to having to tweak the index file to control the injection of material-ui css (due to a specificity clash between my css and material-ui/icons css)
- 2) Thought the above was too complicated, so decided to use the material svg markup directly and get rid of @material-ui/icons for this component.
- 3) Removed the button, because why have a button inside an anchor. This resulted in getting to remove most of the css. Now the css file with just a single rule seems almost silly, but I do like using css modules rather than the inline styles.

Let me know what you think.

amcgee:
@jenniferarnesen looking now!

amcgee (approved):
Looks great, and very clean!

One possible enhancement would be to extract the SVG into a separate file so that it's self-contained and could potentially be re-used.

PR #307, opened 2019-05-03
chore: replace old material-ui IconButton

jenniferarnesen:
To view and try the button:
Open a dashboard in edit view. Click to add the "DHIS 2 Home Page" resource. A new item should be added. In that item, there are 2 red trash bin icons: one for the whole item, and one for the "Home Page" in the item. It is the second red trash bin that this PR has changed.


neeilya (approved):
Tested the code, works as expected.

PR #308, opened 2019-05-07
fix(styles): fix styles for filter selector

edoardo:
Open the filter dialogs within the app container for inheriting the app's
styles.
Fixes DHIS2-6804.

Requires an updated @dhis2/d2-ui-analytics.

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,7 +6,7 @@
     "license": "BSD-3-Clause",
     "dependencies": {
         "@dhis2/d2-i18n": "^1.0.4",
-        "@dhis2/d2-ui-analytics": "^1.0.2",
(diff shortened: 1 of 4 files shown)


janhenrikoverland (commented):
Would it work if we e.g. passed in a component instead to avoid the global id?

janhenrikoverland (approved):

edoardo:
Found a problem: the page can be scrolled under the open DimensionsPanel.

edoardo:
Ended up using a different solution.
The modal is still attached to the `<body>` as normal, so the scrolling issue is not happening.
`font-family` is applied to the Popover component.

janhenrikoverland (approved):

PR #309, opened 2019-05-07
chore: replace old mui edit buttons and dialog in dashboard edit view

jenniferarnesen:
Use latest material-ui, or ui-core where applicable

Changes affect edit mode of dashboard:
- Save changes, Delete, Translate and Exit without saving buttons
- Confirm delete dialog


cooper-joe:
Looks good! I have one request, and that is to switch the order of the translate/delete buttons. So, button order should be Save changes, Translate, Delete.

cooper-joe (changes requested):
Looks good! I have one request, and that is to switch the order of the translate/delete buttons. So, button order should be Save changes, Translate, Delete.

edoardo (approved):
LGTM

cooper-joe (approved):

jenniferarnesen:
Use latest material-ui, or ui-core where applicable

Changes affect edit mode of dashboard:

* Save changes, Delete, Translate and Exit without saving buttons
* Confirm delete dialog

PR #313, opened 2019-05-10
chore: replace old mui dashboard item chips with ui-core chips

jenniferarnesen:
image

image

- there were some height adjustments because it seems that the ui-core chips are a few pixels taller than the mui chips

- testing: remember to also adjust the height of the control bar to make sure the chips aren't getting cut-off along the bottom


cooper-joe (changes requested):
Ideally we could define the color of the icon when chips are unselected? I think grey700 would be a little nicer :)

jenniferarnesen:
with grey700:

image

cooper-joe (approved):

janhenrikoverland (approved):

PR #314, opened 2019-05-13
chore: rich text fix parsing of emphasis in links [v32]

edoardo:


jenniferarnesen:
Could you change the title and description to state what is changing from the user's perspective? (e.g., added support for xy, or fixed bug yz)

jenniferarnesen:
Also, do you know why both 5.3.11 and ^5.3.11 are in the yarn.lock file?

edoardo:
> Also, do you know why both 5.3.11 and ^5.3.11 are in the yarn.lock file?

After updating package.json and removing node_modules i run `yarn install`.

jenniferarnesen (approved):

PR #316, opened 2019-05-13
chore: use new material-ui instead of d2-ui-core for TextField

jenniferarnesen:
d2-ui-core TextField uses old material-ui so replace it with direct use of new material-ui.
This will again be replaced when ui-core textarea is published

### DIFF
--- a/src/components/Item/TextItem/Item.js
+++ b/src/components/Item/TextItem/Item.js
@@ -64,8 +64,6 @@ const TextItem = props => {
                         <Input
                             value={text}
                             multiline
-                            rows={1}
-                            rowsMax={15}
                             fullWidth
                             style={style.textField}
                             placeholder={i18n.t('Add text here')}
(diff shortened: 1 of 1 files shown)


src/components/Item/TextItem/Item.js
```diff
@@ -60,10 +61,11 @@ const TextItem = props => {
                 <Line />
                 <div className="dashboard-item-content">
                     <RichTextEditor onEdit={onChangeText}>
-                        <TextField
+                        <Input
                             value={text}
                             multiline
                             rows={1}
+                            rowsMax={15}
```
edoardo: Any reason for adding this? I had to remove it for fixing another bug, see: https://github.com/dhis2/dashboards-app/commit/9419d281f477ce61b75352f1d1ef7ef1c152b0a3.

edoardo (commented):
LGTM.

src/components/Item/TextItem/Item.js
jenniferarnesen: Right - I knew there was something. Fixed it by removing both the `rows` and `rowsMax` properties.

edoardo (approved):

PR #319, opened 2019-05-20
chore: replace mui dialog with ui-core modal

jenniferarnesen:
Also, use ui-core CssReset instead of importing old reset.css file


jenniferarnesen:
This PR is WIP because I'm waiting for an update to the ui-core Modal component that will set the spacing between the action buttons

PR #326, opened 2019-06-04
fix: conditionally show the Show More button in item selector

jenniferarnesen:
The button should only be shown if there are more items that can be fetched (e.g., more than the 5 displayed items).  The way this is done is to fetch 6 items even though only 5 will be displayed. Then we can easily check if 6 items were returned in the api response.

There is also a minor amount of code clean-up - removing unneeded properties, moving a function to where it actually belongs.

Fixes [https://jira.dhis2.org/browse/DHIS2-7076]


neeilya (approved):
Tested - works as expected now 👍

PR #336, opened 2019-09-24
fix: hide vis action buttons for single value charts

janhenrikoverland:
Separates interpretation and vis action buttons inside the VisualizationItemHeaderButtons component to be able to conditionally render only the interpretation button (hiding the vis buttons for single value charts).

Also fixes the problem where the bottom pixel row of the AO title gets cut off.

### DIFF
--- a/src/components/Item/VisualizationItem/Item.js
+++ b/src/components/Item/VisualizationItem/Item.js
@@ -254,21 +254,20 @@ export class Item extends Component {
         );
     };
 
-    getActionButtons = () => {
-        const { item, visualization } = this.props;
-
-        return pluginManager.pluginIsAvailable(item, visualization) &&
-            !this.props.editMode ? (
+    getActionButtons = () =>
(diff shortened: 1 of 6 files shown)


src/components/Item/VisualizationItem/Item.js
```diff
@@ -229,7 +229,7 @@ export class Item extends Component {
         const itemName = pluginManager.getName(item);
 
         return (
-            <div style={{ display: 'flex', alignItems: 'center' }}>
+            <div style={{ display: 'flex' }}>
```
edoardo, line 232: Is this on purpose?
I think I added it for centering SV within the dashboard item.

edoardo (approved):
Just the one comment, other than that it looks fine.

src/components/Item/VisualizationItem/Item.js
janhenrikoverland: From what I can see it only affects the title row.

PR #348, opened 2019-10-09
fix: upgrade packages + convert to function components with hooks

janhenrikoverland:
- Upgrade packages
- Convert class components to function components and reimplement lifecycles with useState and useEffect hooks (deprecation warnings in the console)

### DIFF
--- a/src/components/ControlBar/Filter.js
+++ b/src/components/ControlBar/Filter.js
@@ -1,4 +1,4 @@
-import React, { Component } from 'react';
+import React, { useState, useEffect } from 'react';
 import PropTypes from 'prop-types';
 import { withStyles } from '@material-ui/core/styles';
 import i18n from '@dhis2/d2-i18n';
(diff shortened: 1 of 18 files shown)


edoardo (approved):
Looks good, but i was expecting to see `package.json` modified too.

src/components/ControlBar/Filter.js
```diff
@@ -27,69 +27,57 @@ const styles = {
...
+export const Filter = props => {
+    const [value, setValue] = useState(DEFAULT_STATE_DASHBOARDS_FILTER_NAME);
 
-        this.state = {
-            value: DEFAULT_STATE_DASHBOARDS_FILTER_NAME,
-        };
-    }
+    useEffect(() => {
+        setValue(props.name);
+    }, [props.name]);
```
amcgee, line 35: This actually has a slightly different behavior than the previous version - before, the "state" would always start as DEFAULT_STATE_DASHBOARDS_FILTER_NAME even if `props.name` was set; after, it will render once as `DEFAULT_STATE_DASHBOARDS_FILTER_NAME` and then immediately re-render as `props.name`

Since `setValue` is only ever called in `useEffect`, I actually don't think the hooks are necessary - it could just be `const value = props.name || DEFAULT_STATE_DASHBOARDS_FILTER_NAME`

**NOTE** This would actually cause the "escape" behavior to reset to the default, rather than to blank, is that something we want to avoid?

src/components/ControlBar/Filter.js
```diff
@@ -27,69 +27,57 @@ const styles = {
...
-    setFilterValue = event => {
+    const setFilterValue = event => {
         event.preventDefault();
 
-        this.props.onChangeName(event.target.value);
+        props.onChangeName(event.target.value);
     };
 
-    handleKeyUp = event => {
+    const handleKeyUp = event => {
```
amcgee, line 43: To make sure this is a stable function reference (better for pure components) this could be `useCallback(event => {...}, [props])`, otherwise it will recreate the function on every render (unlike the previous version)

src/components/ControlBar/__tests__/Filter.spec.js
```diff
@@ -27,18 +27,6 @@ describe('Filter', () => {
         expect(filter()).toMatchSnapshot();
     });
 
-    describe('when updated name property is provided', () => {
```
amcgee, line 30: Why was this test removed?

src/components/ItemGrid/ItemGrid.js
```diff
@@ -44,42 +44,36 @@ import ProgressiveLoadingContainer from '../Item/ProgressiveLoadingContainer';
...
-export class ItemGrid extends Component {
-    state = {
-        expandedItems: {},
-    };
+export const ItemGrid = props => {
+    const [expandedItems, setExpandedItems] = useState({});
+
+    useEffect(() => {
+        props.edit && setExpandedItems({});
+    }, [props.edit]);
```
amcgee, line 52: I **think** (worth confirming) that using sub-properties (`props.edit`) in the dependency array won't work as expected, since the `props` variable is the one bound to the function initialization.  So I think it would need to be `[props]` as dependency.

amcgee (changes requested):
Looks way cleaner!  Left a few comments - we might want to use `useCallback` instead of regenerating callback functions on each render.

src/components/ControlBar/Filter.js
janhenrikoverland: Agreed (now that I know useCallback :-)

src/components/ControlBar/Filter.js
janhenrikoverland: You are right, removing the hooks is actually better.

Regarding "escape", the redux reducer will return default when the component calls props.onChangeName() without arguments.

amcgee:
Oof, this got stale but it's still pretty great.  Can we take this `off-hold`?

jenniferarnesen:
Superceded by this PR: https://github.com/dhis2/dashboards-app/pull/1177

PR #519, opened 2019-12-11
refactor: eslint errors

martinkrulltott:
* Implements the dhis2/cli-style ES Lint rules and fixes 200+ errors.


janhenrikoverland (approved):

janhenrikoverland (approved):

package.json
```diff
@@ -43,7 +43,7 @@
         "analyze": "source-map-explorer 'build/static/js/*.js'",
         "prestart": "npm run extract-pot && npm run localize && npm run manifest",
         "start": "react-scripts start",
-        "lint": "eslint -c .eslintrc.json src",
+        "lint": "eslint -c .eslintrc.js src",
```
varl, line 46: Why not run `d2-style js check` here (or `apply`)?

Or are you avoiding the prettier config?

martinkrulltott: Valid point, but haven't considered it for this fix, as the scope was only to fix the Travis issue caused by using `.js` instead of `.json`.

varl: OK. I see this repo is not formatted with the frontend Prettier config, so there are going to be massive changes triggered.

Seems wise to do it in a different commit at the very least.

martinkrulltott: New branch/PR with suggested change coming up 🙂 🚀

PR #520, opened 2019-12-12
fix: cli-style commit hooks

martinkrulltott:
🎉 Adds commit hooks using [cli-style](https://github.com/dhis2/cli-style) and Husky 💻 🚀 

🔨 Before commit, the following will be run and validated:
1. JS code using `d2-style js check --staged`
1. Text using `d2-style text check --staged`
1. Commit message

🔨 Before push, the following will be run and validated:
1. Test suite

💡  A _dry run_ of the commit validation can be run using `yarn validate-commit`.
💡  A _dry run_ of the push validation can be run using `yarn validate-push`.

### DIFF
--- a/.huskyrc.js
+++ b/.huskyrc.js
@@ -1,0 +1,11 @@
+const { config } = require('@dhis2/cli-style');
+const husky = require(config.husky);
+
+const tasks = arr => arr.join(' && ');
+
+module.exports = {
+    hooks: {
+        ...husky.hooks,
+        'pre-commit': tasks(['yarn validate-commit']),
+    },
+};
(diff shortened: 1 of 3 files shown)


.huskyrc.js
```diff
@@ -0,0 +1,11 @@
+const { config } = require('@dhis2/cli-style');
+const husky = require(config.husky);
+
+const tasks = arr => arr.join(' && ');
+
+module.exports = {
+    hooks: {
+        ...husky.hooks,
+        'pre-commit': tasks(['yarn validate-commit']),
```
varl: Technically you don't need the `tasks` function here if you only want a single command.

varl (approved):

janhenrikoverland (approved):

PR #523, opened 2020-01-03
fix: make dashboard items draggable again by passing down event props

janhenrikoverland:
Fixes https://jira.dhis2.org/browse/DHIS2-8091.

- Not destructuring "debounceMs" and "bufferFactor" from props to avoid unused variables
- Picking only the event props from props to avoid passing the above props to the dom element


amcgee (approved):
LGTM - we're sure there are no other props that could be passed and that we'd want to forward on to the DOM element?

janhenrikoverland:
@amcgee We could instead take out the ones we don't want and pass on the rest, but blindly passing on React component props to a dom element is maybe too liberal. So we would have to maintain it either way I guess.

PR #529, opened 2020-02-06
[WIP] feat: support the visualization type and use new pivot table plugin

jenniferarnesen:


src/icons/VisTypes.js
```diff
@@ -0,0 +1,883 @@
...
+    VIS_TYPE_PIVOT_TABLE,
+} from '@dhis2/analytics';
+
+
+
+const VisIcon = ({ name, style }) => {
+    switch (name) {
+        case VIS_TYPE_COLUMN:
+            return (
+                <svg
```
edoardo, line 27: Can these icon definitions also be imported from `@dhis2/analytics`?
We now have the same icons in several places, including d2-ui `FavoritesDialog` component.

src/modules/itemTypes.js
```diff
@@ -57,16 +57,27 @@ export const getPlugin = type => {
...
 // Item type map
 export const itemTypeMap = {
+    [VISUALIZATION]: {
+        id: VISUALIZATION,
+        endPointName: 'visualizations',
+        propName: 'visualization',
+        pluralTitle: i18n.t('Visualizations'),
+        plugin: 'visualizationPlugin',
+        domainType: DOMAIN_TYPE_AGGREGATE,
+        visualizationType: VISUALIZATION_TYPE_TABLE,
```
edoardo, line 67: I believe this is used in the item header button for switching between chart/pt/map.
Do we still need that functionality for charts/PT? Now we would need to change visType within the vis object passed to the plugin.
Also, the icon here cannot be "fixed" it depends on the current visType of the dashboard item.

edoardo (commented):
Overall looks good.
I noticed a couple of things which we might want to address before merging, especially the one related to `VISUALIZATION_TYPE_(TABLE|CHART)`

jenniferarnesen:
@edoardo Apologies, I should have marked this as WIP. It wasn't quite ready for review.

src/icons/VisTypes.js
jenniferarnesen: Yepp!

PR #539, opened 2020-02-12
chore: use new visualization plugin for report table item types

jenniferarnesen:
The data-visualizer-plugin now renders pivot tables, so use this plugin rather than the legacy reportTablePlugin. Some code was simplified and reorganized in the process.

Fixes [DHIS2-8095]

Some refactorings:
* move visualization actions to new actions/visualizations.js
* renamed actions from "RECEIVED..." to "ADD..." to improve consistency
* simplified code in ItemHeaderButtons


janhenrikoverland (approved):
LGTM

PR #540, opened 2020-02-13
feat: use the new visualizations type instead of report tables and charts

jenniferarnesen:
The api now supports a `visualizations` endpoint, which includes both report tables and charts, and the data visualizer now also supports a table as a type of visualization. Now this concept of merging the two types is coming to Dashboards.

* show "Visualizations" instead of Report Table and Chart in the dashboard item selector
* show an icon specific to the type of chart/table for visualization


janhenrikoverland (approved):
LGTM

PR #541, opened 2020-02-14
fix: interpretation comments delete DHIS2-8299

edoardo:
The interpretations component has been updated to fix the issue with
deleting interpretation comments.

Fixes https://jira.dhis2.org/browse/DHIS2-8299

Bumped the versions for all d2-ui packages, to avoid ending up with different versions of sub-dependencies (assuming that sub-dependencies use the same version within d2-ui packages).

Link to the PR in d2-ui with the actual fix: https://github.com/dhis2/d2-ui/pull/543


janhenrikoverland (approved):

martinkrulltott (approved):
LGTM but the build is failing and there's a minor conflict that needs to be resolved

PR #567, opened 2020-02-21
fix: filter width + input bug

martinkrulltott:
* Sets the filter popup to the correct width (320px, confirmed by @cooper-joe) 
* Fixes a bug where the filter popup would change width when the user starts to type in the filter input field (attached video)

Depends on https://github.com/dhis2/analytics/pull/316

Bug:
ezgif com-video-to-gif

Correct width:
image

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,7 +5,7 @@
     "private": true,
     "license": "BSD-3-Clause",
     "dependencies": {
-        "@dhis2/analytics": "^4.0.2",
+        "@dhis2/analytics": "^4.1.5",
         "@dhis2/app-runtime": "^2.0.3",
         "@dhis2/d2-i18n": "^1.0.6",
         "@dhis2/d2-ui-core": "^6.5.8",
--- a/src/components/ItemFilter/FilterSelector.js
+++ b/src/components/ItemFilter/FilterSelector.js
@@ -132,6 +132,7 @@ class FilterSelector extends Component {
                     }}
                 >
                     <DimensionsPanel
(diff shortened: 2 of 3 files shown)


janhenrikoverland (approved):

jenniferarnesen (approved):

martinkrulltott:
As pointed out by @jenniferarnesen , this affects the wrapping of long names. They wrap, but are too narrow.
image

Previously, DV dimensions panel (_DP_) and Dashboards DP used to be the same width, which is why this was still working. But since Dashboards DP will now deviate style wise from DV and they're both separated (since the introduction of React Beautiful DnD), can we safely change the width for the DP in Analytics directly? (and fix the wrapping issue at the same time) Rather than passing in a style prop? @janhenrikoverland

jenniferarnesen:
I would say yes. The one in analytics seems to be a custom one for the dashboards now.

janhenrikoverland:
@martinkrulltott @jenniferarnesen Or even move it to dashboards-app?

martinkrulltott:
As the issue was in DimensionItem (thanks for pointing this out @jenniferarnesen ), moving DimensionsPanel wouldn't fix this anyway. Instead I updated DimensionItem (https://github.com/dhis2/analytics/pull/316/commits/ee4eef89b5586c1365f5482d5a411dd594dd36ed) to work in both use cases. Using a hardcoded max-width was a poor implementation anyway (whoopsi), so this solution is better in both regards.

jenniferarnesen (approved):

PR #568, opened 2020-02-24
feat: condense item header options into a single menu

jenniferarnesen:
Single button to access all the options
image

Here are the options for a typical chart. Note that Single value and YoY types won't have the "View as" options
image

As part of the implementation, the DeleteItemButton was moved to the ItemHeader, rather than being absolutely positioned in the ItemGrid.

Implements [[DHIS2-5822]](https://jira.dhis2.org/browse/DHIS2-5822)


src/components/Item/NotSupportedItem/Item.js
```diff
@@ -5,7 +5,10 @@ import NotInterestedIcon from '@material-ui/icons/NotInterested';
 
 const NotSupportedItem = props => (
     <Fragment>
-        <ItemHeader title={`Item type not supported: ${props.item.type}`} />
+        <ItemHeader
+            title={`Item type not supported: ${props.item.type}`}
```
martinkrulltott: i18n for this string?

src/components/Item/VisualizationItem/ItemHeaderButtons.js
```diff
@@ -15,178 +20,127 @@ import {
...
+            handleClose();
         }
+    };
+
+    const handleClose = () => setAnchorEl(null);
+
+    const type = visualization.type || item.type;
+    const canViewAs =
+        !isSingleValue(props.visualization.type) &&
+        !type.match(/^YEAR_OVER_YEAR/);
```
martinkrulltott: Could we use the Analytics fn `isYearOverYear` instead?

src/components/Item/VisualizationItem/ItemHeaderButtons.js
```diff
@@ -15,178 +20,127 @@ import {
...
+                    {canViewAs && (
+                        <>
+                            <ViewAsMenuItems />
+                            <Divider />
+                        </>
+                    )}
+                    <MenuItem
+                        dense
+                        icon={<LaunchIcon style={{ fill: '#6e7a8a' }} />}
+                        label={i18n.t(`Open in ${getAppName(item.type)} app`)}
```
martinkrulltott: Incorrect i18n [interpolation](https://www.i18next.com/translation-function/interpolation)? Example of our standard implementation here: https://github.com/dhis2/dashboards-app/blob/6b7ba3f599d46308c9dbc02c0d5f43aecf7b4bb9/src/components/FilterBar/FilterBadge.js#L44

martinkrulltott: Also, I'm surprised that this doesn't show up as a change in https://github.com/dhis2/dashboards-app/blob/6b7ba3f599d46308c9dbc02c0d5f43aecf7b4bb9/i18n/en.pot, maybe it needs to be regenerated?

src/components/Item/VisualizationItem/__tests__/ItemHeaderButtons.spec.js
```diff
@@ -1,24 +0,0 @@
-import React from 'react';
```
martinkrulltott: Why delete this test file completely? `ItemHeaderButtons` is still used

martinkrulltott (commented):
Reviewed with comments. Otherwise the code LGTM. Haven't pulled and tested though. Could you provide screenshots of what it looks like after the fix?

src/components/Item/VisualizationItem/ItemHeaderButtons.js
```diff
@@ -15,178 +20,127 @@ import {
...
+
+    const type = visualization.type || item.type;
+    const canViewAs =
+        !isSingleValue(props.visualization.type) &&
+        !type.match(/^YEAR_OVER_YEAR/);
+
+    const interpretationMenuLabel = props.activeFooter
+        ? i18n.t(`Show interpretations and details`)
+        : i18n.t(`View interpretations and details`);
+
```
cooper-joe, line 69: These messages are mixed up and one of them needs to be made to it's opposite. Right now when the interpretations are visible the menu item says 'Show interpretations and details'. This should be 'Hide interpretations and details'.

src/components/Item/styles/ItemHeader.module.css
```diff
@@ -0,0 +1,28 @@
+.itemWrap {
+    padding: var(--spacers-dp8);
+}
+
+.itemHeaderWrap {
+    display: flex;
+    align-items: top;
+    margin: var(--spacers-dp4) var(--spacers-dp4) var(--spacers-dp8)
+        var(--spacers-dp4);
```
cooper-joe: Could this be adjusted to: `margin: var(--spacers-dp4) var(--spacers-dp4) var(--spacers-dp8) var(--spacers-dp8);`

src/components/Item/VisualizationItem/ItemHeaderButtons.js
cooper-joe: I'm not sure where this is defined, but could we use the string 'Open in Data Visualizer app' rather than 'Open in Visualizer app...'?

src/components/Item/ItemHeader.js
```diff
@@ -1,28 +1,59 @@
...
-                style={{ userSelect: editMode ? 'none' : 'text' }}
-            >
-                {title}
-            </div>
-            {actionButtons}
+        <div className={classes.itemHeaderWrap}>
+            <p className={classes.itemTitle}>{title}</p>
+            {editMode ? (
+                <div className={classes.itemActionsWrap}>
+                    <DeleteItemButton onClick={handleDeleteItem} />
```
cooper-joe, line 27: Is it possible to add a `title` attribute to this `button`, just until our Tooltip components are ready? `title="Delete item"` would help an unsure user understand this icon on hover.

cooper-joe (changes requested):
Looking good, much cleaner with multiline titles! Thanks for making the changes to the delete button in edit mode, that works well. I've requested a few minor changes to styles/labels.

src/components/Item/ItemHeader.js
martinkrulltott: `title="Delete item" ` + i18n 🙂🚀

src/components/Item/VisualizationItem/ItemHeaderButtons.js
cooper-joe: After reading your JIRA message, I agree. Let's adjust to:
`Show interpretations and details` and `Hide interpretations and details`

src/components/Item/VisualizationItem/ItemHeaderButtons.js
jenniferarnesen: absolutely. Sometimes when moving code, you kind of miss those things :)

src/components/Item/VisualizationItem/__tests__/ItemHeaderButtons.spec.js
jenniferarnesen: You're right. I think I had deleted it earlier in the process when I removed that file.

src/components/Item/VisualizationItem/ItemHeaderButtons.js
jenniferarnesen: oops. Will fix

src/components/Item/styles/ItemHeader.module.css
jenniferarnesen: Yepp.

jenniferarnesen:
image

src/components/Item/NotSupportedItem/Item.js
```diff
@@ -1,11 +1,17 @@
...
 import PropTypes from 'prop-types';
+import i18n from '@dhis2/d2-i18n';
 import ItemHeader from '../ItemHeader';
 import NotInterestedIcon from '@material-ui/icons/NotInterested';
 
 const NotSupportedItem = props => (
     <Fragment>
-        <ItemHeader title={`Item type not supported: ${props.item.type}`} />
+        <ItemHeader
+            title={i18n.t('Item type "{{type}}" not supported', {
```
martinkrulltott, line 10: Are the double-quotes around {{type}} intentional? They will show up in the result, as: Item type "SomeType" not supported

src/components/Item/VisualizationItem/ItemHeaderButtons.js
```diff
@@ -15,178 +20,127 @@ import {
...
+        props.onToggleFooter();
+        if (anchorEl !== null) {
+            handleClose();
         }
+    };
+
+    const handleClose = () => setAnchorEl(null);
+
+    const type = visualization.type || item.type;
+    const canViewAs = !isSingleValue(type) && !isYearOverYear(type);
```
martinkrulltott, line 64: Tiny detail, but what does `canViewAs` mean? Is "ViewAs" a verb? Further down it seems to be used as a noun for the component `ViewAsMenuItems` 🤷‍♂

martinkrulltott (approved):
LGTM, two tiny comments

cooper-joe (approved):
Thanks for making those changes, LGTM.

src/components/Item/NotSupportedItem/Item.js
jenniferarnesen: That was intentional, yes.

src/components/Item/VisualizationItem/ItemHeaderButtons.js
jenniferarnesen: Maybe would have been better with a more descriptive name, like "canViewAsOtherTypes", "hasOtherViews".

PR #592, opened 2020-02-28
WIP: fix: do not rerender visualizations unless they have actually changed

jenniferarnesen:
Fixes issue where all charts and tables on the dashboard were rerendering when show/hiding interpretations for one item, or when draggin/resizing an item in edit mode.

A likely reason for this is the useEffect hook in the Visualizer plugin, where, most likely there is a shallow compare of properties (i.e., the visualization property). This object was being considered different than the previous one, so the visualization was rerendered.

The solution here is to store the current view in the redux store, so that the same object is passed to the plugin, unless the view has changed (e.g. toggling to chart/map).  The store now looks a little more like the DV store, in that there is the original visualization, and there is the current view of the visualization.

### DIFF
--- a/src/components/Item/VisualizationItem/Item.js
+++ b/src/components/Item/VisualizationItem/Item.js
@@ -144,12 +144,6 @@ export class Item extends Component {
         }
         const style = this.getContentStyle();
 
-        // console.log(
-        //     'current',
-        //     currentVisualizationView.name,
-        //     currentVisualizationView.type
-        // );
-
         switch (this.getActiveType()) {
             case VISUALIZATION:
             case CHART:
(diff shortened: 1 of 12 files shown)


src/components/Item/VisualizationItem/Item.js
```diff
@@ -119,81 +124,82 @@ export class Item extends Component {
...
-
-        switch (activeType) {
+        switch (this.getActiveType()) {
             case VISUALIZATION:
             case CHART:
             case REPORT_TABLE: {
                 return (
                     <VisualizationPlugin
                         d2={this.d2}
                         visualization={applyFilters(
```
edoardo, line 152: Because the filtering is done here, the `currentVisualizationView` won't match the object passed to the plugin, causing a rerender of all visualization items (since the filter is applied to all items).
`currentVisualizationView` should contain the filtered visualization, and change every time either `visualization` or `itemFilters` change.

PR #612, opened 2020-03-04
fix: clear the filters after the dashboard has been switched

jenniferarnesen:
Bug: when switching from one dashboard to another, and item filters were set, then dashboard items from the first dashboard were being re-rendered briefly before the new dashboard was loaded.  This was happening because the item filters were being cleared from the store prior to the dashboard switching. The store change triggered a re-render on the dashboard items.

As part of the fix, several imports were added to `actions/selected`. I reordered the imports to group them, just for reading legibility.


martinkrulltott (approved):
LGTM

PR #616, opened 2020-03-06
fix: hide current vis type as option in 'view as' (DHIS2-8400)

martinkrulltott:
Previously the options for 'View as' include the current type of the visualization. As an example, you can chose to view a chart as a chart. https://jira.dhis2.org/browse/DHIS2-8400

- Prevents the current vis type to be displayed as an option in the 'View as' menu
- Makes sure that the menu is closed as soon as an option is clicked, to prevent a flash of ghost items (i.e. the options would change just before the menu had time to close)

-----------

Examples:

Map:

image

Chart:

image

Pivot table:

image

Event report:

image


src/components/Item/VisualizationItem/ItemHeaderButtons.js
```diff
@@ -78,24 +78,25 @@ const ItemHeaderButtons = props => {
...
-            />
-            <MenuItem
-                dense
-                active={[REPORT_TABLE, EVENT_REPORT].includes(activeType)}
-                label={i18n.t('View as Table')}
-                onClick={onViewTable}
-                icon={<TableIcon style={iconFill} />}
-            />
-            {itemHasMapView() && (
+            {activeType !== CHART && (
```
martinkrulltott: Should this include `EVENT_CHART` as well?

martinkrulltott: Should this include `EVENT_CHART` as well? @janhenrikoverland @jenniferarnesen

jenniferarnesen: Yes

jenniferarnesen (approved):

src/components/Item/VisualizationItem/ItemHeaderButtons.js
jenniferarnesen: Fixed it.

martinkrulltott: Thanks!

PR #617, opened 2020-03-09
fix: edit mode performance improvements

janhenrikoverland:
- Uses memoization in ItemGrid.js on dashboard items to avoid unnecessary rerenders
- Changes the RECEIVED_DASHBOARD_LAYOUT reducer to not always produce new dashboard items (important). Also simply returns current state when no items changed.


src/components/Item/VisualizationItem/Item.js
```diff
@@ -77,11 +77,9 @@ export class Item extends Component {
 
         this.contentRef = React.createRef();
 
-        this.memoizedApplyFilters = memoizeOne(this.applyFilters);
+        this.memoizedApplyFilters = memoize(this.applyFilters);
```
jenniferarnesen: Using lodash memoize breaks the filtering (see the documentation - only the first value of the memoized function is used as the cache key).  So I think we should stick with memoizeOne in this file.

src/components/ItemGrid/ItemGrid.js
```diff
@@ -49,6 +50,15 @@ export class ItemGrid extends Component {
         expandedItems: {},
     };
 
+    constructor(props) {
+        super(props);
+
+        this.getMemoizedItem = memoize(this.getItem);
+        this.getMemoizedItems = memoize(this.getItems);
+        this.getMemoizedItemComponent = memoize(this.getItemComponent);
+        this.getMemoizedItemComponents = memoize(this.getItemComponents);
```
jenniferarnesen: I tried using each of these on their own. The only one that actually decreased number of renders was getMemoizedItem (line 56). So I think we can safely remove the other 3 and have the same result.

jenniferarnesen (changes requested):
Edit mode is so much faster! See comments for requested changes.

janhenrikoverland:
Typing in a textbox is fast again as it doesn't affect other items anymore. Same thing for resizing when it doesn't touch other (or only a few) items. However resizing one of the top items vertically which pushes all the other items is still a bit slow (was way faster in 2.33), need to investigate further on this.

PR #618, opened 2020-03-10
fix: set font-weight to 700 according to spec

janhenrikoverland:
Change the dashboard item header title font weight from 500 to 700 according to the spec.

Before:
Screenshot from 2020-03-10 11-45-54

After:
Screenshot from 2020-03-10 11-45-37


martinkrulltott (approved):
Yes! 🎉

PR #621, opened 2020-03-11
fix: wrap VisualizationPlugin with fatal error boundary

jenniferarnesen:
Crashes in the VisualizationPlugin will be caught, so that the only the affected dashboard item doesn't display, rather than bringing down the whole dashboard app.

image

In the calculation of the height style prop, it was necessary to check for existence of `contentRef` since it doesn't exist when there's been an error.

I also found a bug. If I click on interpretations,  The item height calculation is incorrect. It is a visual bug. Not sure if this is a showstopper. I can continue looking at it, but probably won't solve it by the deadline.

image

### DIFF
--- a/i18n/en.pot
+++ b/i18n/en.pot
@@ -5,8 +5,8 @@ msgstr ""
 "Content-Type: text/plain; charset=utf-8\n"
 "Content-Transfer-Encoding: 8bit\n"
 "Plural-Forms: nplurals=2; plural=(n != 1)\n"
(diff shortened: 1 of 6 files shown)


amcgee (approved):
Didn't test but the error boundary functionality looks good - a bit of styling love would be good but not a huge deal, since the main point of this change is to prevent a dashboard item from crashing the entire dashboard.  If we want to make small visual changes or fix the interpretations bug after the freeze I think that's fine.

PR #664, opened 2020-03-18
fix: calculate item height accounting for long title [DHIS2-8492]

jenniferarnesen:
fixes: https://jira.dhis2.org/browse/DHIS2-8492

Calculate the height of the item based on the header height - which is no longer constant since items now support wrapped titles.

This calculation is the backup value incase the measured height of the chart is not available right away, as in view mode.

In addition, fn getContentStyle was memoized to reduce chart rerenders when items are moved in edit mode.

Before (notice the chart legend is not shown at the bottom):

image

Fixed:

image


edoardo (approved):
LGTM

PR #674, opened 2020-03-25
fix: use measured height in view mode [DHIS2-8492]

jenniferarnesen:
Fixes issue where reloading a dashboard can cause a map item to cut off the bottom of the map. This may happen if you open the dashboard, then click on the same dashboard item chip to reload the dashboard. Note that the bottom of the map is clipped in the screenshot:

image

The Jira ticket has a video of the bug.

### DIFF
--- a/src/components/Item/VisualizationItem/Item.js
+++ b/src/components/Item/VisualizationItem/Item.js
@@ -169,13 +169,18 @@ export class Item extends Component {
             );
         }
 
+        const calculatedHeight =
+            this.props.item.originalHeight -
+            this.headerRef.current.clientHeight -
+            HEADER_MARGIN_HEIGHT;
+
         const props = {
             ...this.props,
             visualization,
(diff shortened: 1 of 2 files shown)


janhenrikoverland:
Clicking the same dashboard again seems to work :+1: 

Still see this though (no padding between background map and item border):
Screenshot from 2020-03-25 12-03-00

Expected:
Screenshot from 2020-03-25 22-11-51

jenniferarnesen:
> Clicking the same dashboard again seems to work 👍
> 
> Still see this though (no padding between background map and item border):
> Screenshot from 2020-03-25 12-03-00
> 
> Expected:
> Screenshot from 2020-03-25 22-11-51

Good eye. Fixed. The height didn't take into account padding on the dashboard-item-content.

janhenrikoverland (approved):
LG

PR #677, opened 2020-03-26
chore: upgrade react and some other dependencies [TECH-336]

jenniferarnesen:
No functional changes.

* ItemHeader needed a small refactor due to changes in react-redux
* Removed use of theme, which is for mui, which is going away. (LoadingMask, FilterSelector)
* css tweaks for the Item Selector, needed due to the ui-core upgrade

material-ui/core was not upgraded since the upgrade caused issues in the Translation and Sharing dialog.  Will need to address that later.

https://jira.dhis2.org/browse/TECH-336


martinkrulltott (commented):
The code looks good, haven't tested locally, but you're sure this has no visual impact?
Is it safe to upgrade to React v16.13.1? All other apps seems to be running 16.8.x.

jenniferarnesen:
> The code looks good, haven't tested locally, but you're sure this has no visual impact?
> Is it safe to upgrade to React v16.13.1? All other apps seems to be running 16.8.x.

I haven't spotted any visual impacts. It all looks the same.
Per the discussion in the analytics team meeting, it is fine to upgrade to the latest React.

martinkrulltott (approved):
LGTM

Re: the React upgrade. Hopefully this won't cause any issues, since it's only a minor upgrade. Given that the release is as far away as possible I think we'll have plenty of time to test this ad-hoc during development anyway, so I won't be testing this now.


> material-ui/core was not upgraded since the upgrade caused issues in the Translation and Sharing dialog. Will need to address that later.

Agreed, `material-ui/core` should ideally be phased out completely for this release and be replaced by corresponding `ui-core` components.

PR #685, opened 2020-03-27
refactor: consolidate item conversion functionality [TECH-335]

jenniferarnesen:
Functional changes:
* SPACER type is no longer converted to TEXT type when saving, because `type` is never used on the backend and never has been. So there is no point in doing the conversion on the frontend. Backend looks only at the `text` property. (see this backend code: https://github.com/dhis2/dhis2-core/blob/master/dhis-2/dhis-api/src/main/java/org/hisp/dhis/dashboard/DashboardItem.java#L160)also here shows what is actually stored in the db: https://github.com/dhis2/dhis2-core/blob/master/dhis-2/dhis-services/dhis-service-core/src/main/resources/org/hisp/dhis/dashboard/hibernate/DashboardItem.hbm.xml

* When loading dashboard items, no longer convert VISUALIZATION to CHART/REPORT_TABLE, since the ui supports VISUALIZATION. (there was a comment in the old code pointing this out)

Refactor changes:
* conversion code has been moved and consolidated in a new module. Any constants that were only in use in the conversion code are co-located with the conversion code.

https://jira.dhis2.org/browse/TECH-335

### DIFF
--- a/src/modules/uiBackendItemConverter.js
+++ b/src/modules/uiBackendItemConverter.js
@@ -1,0 +1,38 @@
+import { spacerContent, TEXT, SPACER } from './itemTypes';
+
+const emptyTextItemContent = 'TEXT_ITEM_WITH_NO_CONTENT';
+const isSpacerType = item => item.type === TEXT && item.text === spacerContent;
+
+export const isTextType = item =>
+    item.type === TEXT && item.text !== spacerContent;
+
+export const convertUiItemsToBackend = items => {
+    return items.map(item => {
+        const text = isTextType(item)
+            ? item.text || emptyTextItemContent
+            : null;
+
+        return {
+            ...item,
+            ...(text ? { text } : {}),
+        };
+    });
+};
+
+export const convertBackendItemsToUi = items => {
+    return items.map(item => {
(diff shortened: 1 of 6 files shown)


src/modules/uiBackendItemConverter.js
```diff
@@ -0,0 +1,41 @@
...
+const isUiSpacerType = item => item.type === SPACER;
+const isTextType = item => item.type === TEXT && item.text !== spacerContent;
+
+export const convertUiItemsToBackend = items =>
+    items.map(item => {
+        let text = null;
+        if (isUiSpacerType(item)) {
+            text = spacerContent;
+        } else if (isTextType(item)) {
+            text = item.text || emptyTextItemContent;
```
edoardo, line 17: This would change the actual content stored in the backend and it's different from before.

Are we sure we want the dashboard/dashboardItem APIs to be only consumed by our dashboard app?

jenniferarnesen: Actually, no, there are no changes to what is stored in the backend.

edoardo: 👍 nevermind my comment.

edoardo (approved):

PR #686, opened 2020-03-28
refactor: use css modules for components [TECH-337]

jenniferarnesen:
No functional changes

Changes include:
* replace `<Fragment>` with `<>`
* port css to css modules

Updating the app to follow the css standard determined here: https://github.com/dhis2/notes/issues/79#issuecomment-563933328

Fixes: https://jira.dhis2.org/browse/TECH-337

### DIFF
--- a/src/components/TitleBar/ViewTitleBar.js
+++ b/src/components/TitleBar/ViewTitleBar.js
@@ -1,7 +1,6 @@
-import React, { Component, Fragment } from 'react';
+import React, { Component } from 'react';
 import { connect } from 'react-redux';
 import PropTypes from 'prop-types';
-import { withStyles } from '@material-ui/core/styles';
 import { Link } from 'react-router-dom';
 import i18n from '@dhis2/d2-i18n';
 import SharingDialog from '@dhis2/d2-ui-sharing-dialog';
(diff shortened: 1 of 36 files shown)


src/components/TitleBar/ViewTitleBar.js
```diff
@@ -22,40 +21,10 @@ import {
...
-        alignItems: 'flex-start',
-    },
-    titleBarIcon: {
-        marginLeft: 5,
-        position: 'relative',
-        top: 1,
-        cursor: 'pointer',
-    },
-};
+const NO_DESCRIPTION = i18n.t('No description');
```
martinkrulltott, line 27: Seems like this is only used once, should it still be a const?

src/components/TitleBar/ViewTitleBar.js
```diff
@@ -92,15 +60,15 @@ class ViewTitleBar extends Component {
...
+            <>
                 <div className={classes.titleBar}>
                     <span style={titleStyle}>{name}</span>
                     <div className={classes.actions}>
                         <div
                             className={classes.titleBarIcon}
                             onClick={onStarClick}
                         >
-                            <StarIcon className={classes.starIcon} />
+                            <StarIcon style={{ fill: colors.grey600 }} />
```
martinkrulltott, line 71: className instead of inline styling?

martinkrulltott (approved):
LGTM, just two tiny comments

src/components/TitleBar/ViewTitleBar.js
jenniferarnesen: I had to do this one with the style property due to css priority - MUI was overriding the color. Will get rid of this as soon as we migrate away from MUI icons

src/components/TitleBar/ViewTitleBar.js
jenniferarnesen: Good point

PR #699, opened 2020-04-01
fix: List items not appearing to get added in edit mode (DHIS2-8567)

jenniferarnesen:
The memoized function was not detecting that the Reports (List item) dashboard item had changed if a report was added. Therefore it appeared that the items weren't being added.

Use redux properly rather than resorting to memoize for grid item properties.

Fixes: https://jira.dhis2.org/browse/DHIS2-8567

Thorough testing needed. Memoize was part of a performance improvement (in edit mode) so that needs to be checked again. And we need to make sure all the item types are behaving properly in edit (ie., resize, move, type text, delete, add...) and view mode (i.e., view interpretations).

### DIFF
--- a/src/components/ItemGrid/ItemGrid.js
+++ b/src/components/ItemGrid/ItemGrid.js
@@ -1,7 +1,6 @@
 import React, { Component } from 'react';
 import { connect } from 'react-redux';
 import PropTypes from 'prop-types';
-import memoize from 'lodash/memoize';
 import i18n from '@dhis2/d2-i18n';
 import ReactGridLayout from 'react-grid-layout';
 import { CircularLoader, ScreenCover } from '@dhis2/ui-core';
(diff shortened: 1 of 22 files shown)


martinkrulltott:
Please see comments/discussion in the PR thread on Slack

src/components/ItemGrid/ItemGrid.js
```diff
@@ -100,24 +92,18 @@ export class ItemGrid extends Component {
 
     onRemoveItemWrapper = id => () => this.onRemoveItem(id);
 
-    getItem = dashboardItem => {
+    adjustHeightForExpanded = dashboardItem => {
         const expandedItem = this.state.expandedItems[dashboardItem.id];
-        const hProp = { h: dashboardItem.h };
```
martinkrulltott, line 105: Could be my new favourite line of code 😆 🚀

martinkrulltott (approved):
LGTM, tested locally with all item types and operations that I know of, found a few interesting behaviours (reported on Slack) but nothing that seems related to this PR

PR #713, opened 2020-04-02
Fix/new report items not showing (v34) (DHIS2-8567)

jenniferarnesen:
The memoized function was not detecting that the Reports (List item) dashboard item had changed if a report was added. Therefore it appeared that the items weren't being added.

Use redux properly rather than resorting to memoize for grid item properties.

Fixes: https://jira.dhis2.org/browse/DHIS2-8567

Thorough testing needed. Memoize was part of a performance improvement (in edit mode) so that needs to be checked again. And we need to make sure all the item types are behaving properly in edit (ie., resize, move, type text, delete, add...) and view mode (i.e., view interpretations).


martinkrulltott (approved):
Approved based on my findings from https://github.com/dhis2/dashboards-app/pull/699/. Haven't tested this one locally though.

PR #715, opened 2020-04-08
fix: revert some css module changes and fix missing units

jenniferarnesen:
Fixes: https://jira.dhis2.org/browse/TECH-342
This bug is only in dev, not in 2.34

css modules styles were getting overridden by MUI styles. Also, in css, the units (e.g., "px") must be included.

In screenshots, note the position of "Show more" and the font in the dashboard filter input

Before the fix: 
image

After the fix (should match 2.33 and 2.34):
image


janhenrikoverland (approved):
LG

PR #718, opened 2020-04-09
fix: hide titles for maps (DHIS2-8616)

janhenrikoverland:
Fixes https://jira.dhis2.org/browse/DHIS2-8616 which utilizes https://jira.dhis2.org/browse/DHIS2-8618

Adds support for passing options as part of the plugin call.

Passes `hideTitle: true` to the maps plugin.

Removes this title overlay:
map-with-title


martinkrulltott (approved):
LGTM 🚀

PR #725, opened 2020-04-09
chore: replace mui circular loader with ui-core

jenniferarnesen:
This version also shows the item title and actions, giving the user a little more info and options if a visualization is not loading:

image

[TECH-343]
Fixes: https://jira.dhis2.org/browse/TECH-343

### DIFF
--- a/src/components/Item/VisualizationItem/Item.js
+++ b/src/components/Item/VisualizationItem/Item.js
@@ -162,7 +162,7 @@ export class Item extends Component {
             case REPORT_TABLE: {
                 return (
                     <>
-                        {this.state.pluginIsLoaded && (
+                        {!this.state.pluginIsLoaded && (
                             <div style={props.style}>
                                 <LoadingMask />
                             </div>
--- a/src/components/Item/VisualizationItem/Item.js
+++ b/src/components/Item/VisualizationItem/Item.js
@@ -162,11 +162,11 @@ export class Item extends Component {
             case REPORT_TABLE: {
                 return (
                     <>
-                        {!this.state.pluginIsLoaded ? (
+                        {!this.state.pluginIsLoaded && (
                             <div className={classes.loadingCover}>
                                 <LoadingMask />
                             </div>
-                        ) : null}
+                        )}
                         <VisualizationPlugin
                             d2={this.d2}
                             visualization={this.memoizedApplyFilters(
--- a/src/components/Item/VisualizationItem/Item.js
+++ b/src/components/Item/VisualizationItem/Item.js
@@ -162,8 +162,8 @@ export class Item extends Component {
             case REPORT_TABLE: {
                 return (
                     <>
-                        {!this.state.pluginIsLoaded && (
-                            <div style={props.style} className={classes.cover}>
+                        {this.state.pluginIsLoaded && (
+                            <div style={props.style}>
                                 <LoadingMask />
                             </div>
                         )}
--- a/src/components/Item/VisualizationItem/Item.js
+++ b/src/components/Item/VisualizationItem/Item.js
@@ -163,7 +163,7 @@ export class Item extends Component {
                 return (
                     <>
                         {!this.state.pluginIsLoaded && (
-                            <div className={classes.loadingCover}>
+                            <div style={props.style} className={classes.cover}>
                                 <LoadingMask />
                             </div>
                         )}
--- a/src/components/Item/VisualizationItem/LoadingMask.js
+++ b/src/components/Item/VisualizationItem/LoadingMask.js
@@ -1,13 +1,11 @@
 import React from 'react';
-import { CircularLoader } from '@dhis2/ui-core';
-
-import classes from './styles/LoadingMask.module.css';
+import { ComponentCover, CircularLoader } from '@dhis2/ui-core';
 
 function LoadingMask() {
     return (
-        <div className={classes.mask}>
+        <ComponentCover>
             <CircularLoader />
-        </div>
+        </ComponentCover>
     );
 }
 
--- a/src/components/Item/VisualizationItem/LoadingMask.js
+++ b/src/components/Item/VisualizationItem/LoadingMask.js
@@ -5,10 +5,8 @@ import classes from './styles/LoadingMask.module.css';
 
 const LoadingMask = () => {
     return (
-        <div className={classes.vertical}>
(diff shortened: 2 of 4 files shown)


src/components/Item/VisualizationItem/LoadingMask.js
```diff
@@ -1,14 +1,12 @@
 import React from 'react';
-import CircularProgress from '@material-ui/core/CircularProgress';
+import { ComponentCover, CircularLoader } from '@dhis2/ui-core';
 
-import classes from './styles/LoadingMask.module.css';
-
-function CircularIndeterminate() {
+function LoadingMask() {
```
martinkrulltott: why `function LoadingMask()` over `const LoadingMask = () => {` ?

martinkrulltott (approved):
LGTM. Maybe best to verify with @cooper-joe / @janhenrikoverland that we want the `ComponentCover` for the initial load (gray background) or if we just want to use the plain white background (by just not having the cover) like we do in DV? The cover doesn't "cover" anything, since its just empty space behind (as opposed to an update operation which could cover the old content before displaying the new), but maybe thats how our it was specified in the style guide? (personally I think it would look better with the white bg here, but thats just a matter of taste 🙂 )
image

jenniferarnesen:
@martinkrulltott Yeah, I wasn't sure about the gray background either. So I'll wait for feedback from Joe.

cooper-joe:
> Maybe best to verify with @cooper-joe / @janhenrikoverland that we want the ComponentCover for the initial load (gray background) or if we just want to use the plain white background (by just not having the cover) like we do in DV? The cover doesn't "cover" anything, since its just empty space behind (as opposed to an update operation which could cover the old content before displaying the new), but maybe thats how our it was specified in the style guide? 

@martinkrulltott You are right. The `ComponentCover` is not needed here as there is no interaction to block on first load.

jenniferarnesen:
@martinkrulltott Could you do a new review? I've changed the code quite a bit.

src/components/Item/VisualizationItem/LoadingMask.js
jenniferarnesen: updated

src/components/Item/VisualizationItem/Item.js
```diff
@@ -162,11 +162,11 @@ export class Item extends Component {
             case REPORT_TABLE: {
                 return (
                     <>
-                        {!this.state.pluginIsLoaded ? (
-                            <div className={classes.loadingCover}>
+                        {!this.state.pluginIsLoaded && (
+                            <div style={props.style} className={classes.cover}>
```
martinkrulltott: Is `style={props.style}` needed here? I thought it was used for the `VisualizationPlugin` to calculate the height.

src/components/Item/VisualizationItem/LoadingMask.js
```diff
@@ -1,14 +1,16 @@
...
+import { CircularLoader } from '@dhis2/ui-core';
 
 import classes from './styles/LoadingMask.module.css';
 
-function CircularIndeterminate() {
+const LoadingMask = () => {
     return (
-        <div className={classes.outer}>
-            <CircularProgress className={classes.progress} />
+        <div className={classes.vertical}>
```
martinkrulltott: add `align-items: center` to `vertical` and skip the `horizontal` div?

src/components/Item/VisualizationItem/styles/Item.module.css
```diff
@@ -4,12 +4,12 @@
...
     position: absolute;
-    height: 100%;
-    width: 100%;
+    top: auto;
     left: 0;
-    top: 0;
+    width: 100%;
     z-index: 100;
-    background: #ffffff;
+    background-color: var(--colors-white);
```
martinkrulltott: Skip the `background-color` completely? it's already white by default?

src/components/Item/VisualizationItem/styles/Item.module.css
```diff
@@ -4,12 +4,12 @@
     padding: 10px;
     line-height: 20px;
 }
-.loadingCover {
+
+.cover {
```
martinkrulltott: Actually I think this whole class can be skipped. The width and positioning is handled by the `LoadingMask` and the height is handled dynamically.

src/components/Item/VisualizationItem/Item.js
martinkrulltott: It seems like the `react-grid-item` already has a static height so this could be solved with `display:flex` at the top, `flex-grow: 1` on the `dashboard-item-content` and `height:100%` on further nested children image

martinkrulltott (approved):
Approved with minor comments. Suggestions
- consolidate the `vertical` and `horizontal` divs in to one, with flexbox only.
- remove the `cover` class completely from `Item.js`(and the style file)
- consider to refactor the handling of height, to use css rather than a px value from js

src/components/Item/VisualizationItem/Item.js
jenniferarnesen: To do this, I will need to add a div that wraps the Header and dashboard-item-content. Since it isn't possible to add any css to react-grid-item (it is generated by the React Grid Layout library)

src/components/Item/VisualizationItem/styles/Item.module.css
jenniferarnesen: Yepp you are right. There were so many css changes going on, this was probably some intermediate step that I didn't go back and retest if it was still needed.

src/components/Item/VisualizationItem/LoadingMask.js
jenniferarnesen: Done.

src/components/Item/VisualizationItem/Item.js
jenniferarnesen: Actually, we can add it old-school, by targeting the react-grid-item class directly.

jenniferarnesen: On further investigation, I found a commit that changed how we did this height calculation. Turns out we previously used display: flex as you are recommending, but had to change it in this commit: https://github.com/dhis2/dashboards-app/commit/fab4d387fc9c60528301570e49bf4042b662e625). That was two years ago, and I think it is worth revisiting, but in a separate commit/PR.

PR #726, opened 2020-04-09
chore: replace mui in delete dialog

jenniferarnesen:
No functional changes.
Fixes https://jira.dhis2.org/browse/TECH-338

image

### DIFF
--- a/src/components/ControlBar/styles/ConfirmDeleteDialog.module.css
+++ b/src/components/ControlBar/styles/ConfirmDeleteDialog.module.css
@@ -1,0 +1,4 @@
+.content {
+    font-size: 0.875rem;
+    line-height: 1.5;
+}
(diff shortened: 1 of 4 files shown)


src/components/ControlBar/styles/ConfirmDeleteDialog.module.css
```diff
@@ -0,0 +1,4 @@
+.content {
```
martinkrulltott, line 1: Do we have this style somewhere centralised to be reused? Seems like regular text to me, isn't that already defined by the style guide and implemented somewhere else already?

martinkrulltott (approved):
LGTM. Minor change, I would believe the modal would follow the standard text style that `ui-core` provides. If so, then the `module.css` file could be removed. But if I'm wrong then just keep as is 🚀

src/components/ControlBar/styles/ConfirmDeleteDialog.module.css
jenniferarnesen: Yes, a little strange but I didn't find anything in ui-core. Typography is described in the design docs, but there is nothing implemented in ui-core yet.

Unfortunately there is css from ext.js that is setting the font size and line height which is affecting the modal content, so it had to be overridden, for now. I'll mention it to the ui team and revisit this.

PR #730, opened 2020-04-16
fix: use measured height of dashboard item in view mode [DHIS2-8492] (v34)

jenniferarnesen:
Fixes issue where reloading a dashboard can cause a map item to cut off the bottom of the map. This may happen if you open the dashboard, then click on the same dashboard item chip to reload the dashboard.

Same as this PR https://github.com/dhis2/dashboards-app/pull/674, but for v34


martinkrulltott (approved):
LGTM

PR #700, opened 2020-04-01
fix: add missing reset css for lists (DHIS2-8568)

jenniferarnesen:
fixes: https://jira.dhis2.org/browse/DHIS2-8568

Reset css was removed in this commit: https://github.com/dhis2/dashboards-app/commit/75060637fdd6401948c125d97e9308a27efd8777

PR #731, opened 2020-04-16
fix: add missing reset css for lists (#700) [v34]

jenniferarnesen:
Lists were displaying with bullets, which should have been handled with reset css. Reset css was removed in this commit: 7506063

fixes: https://jira.dhis2.org/browse/DHIS2-8568

Same as this PR, https://github.com/dhis2/dashboards-app/pull/700, but for v34


martinkrulltott (approved):
LGTM. Not tested locally, but trust that it looks the same as https://github.com/dhis2/dashboards-app/pull/700 👍

PR #822, opened 2020-05-13
fix: missing translations v33

jenniferarnesen:
Part of fix for https://jira.dhis2.org/browse/DHIS2-8638

i18n.t must be called at render time, so all calls are placed inside functions. Upgraded analytics and d2-ui components as well, for similar needed fixes.

Most of dashboards translations were already working, but some components were not:

Dashboard filter:

image

Edit mode item selector: 

image

Interpretations

image

Org unit dialog

image


src/components/Item/VisualizationItem/Item.js
```diff
@@ -237,9 +237,9 @@ export class Item extends Component {
                     <a
                         href={pluginManager.getLink(this.props.item, this.d2)}
                         style={{ height: 16 }}
-                        title={`View in ${
-                            itemTypeMap[this.props.item.type].appName
-                        } app`}
+                        title={`View in ${itemTypeMap[
```
martinkrulltott: `i18n`? I don't see this string in [en.pot](https://github.com/dhis2/dashboards-app/blob/868a8e90bc2103efcc9480552e0853ea301971f0/i18n/en.pot)

yarn.lock
```diff
@@ -1022,14 +1022,12 @@
   resolved "https://registry.yarnpkg.com/@csstools/convert-colors/-/convert-colors-1.4.0.tgz#ad495dc41b12e75d588c6db8b9834f08fa131eb7"
   integrity sha512-5a6wqoJV/xEdbRNKVo6I4hO3VjyDq//8q2f9I6PBAvMesJHFauXDorcNCsr9RzvsZnaWi5NYCcfyqP1QeFHFbw==
 
-"@dhis2/analytics@2.4.7":
-  version "2.4.7"
-  resolved "https://registry.yarnpkg.com/@dhis2/analytics/-/analytics-2.4.7.tgz#0cbe4e44db69e690b08a2b284ce08b630142d73f"
-  integrity sha512-mESuLf4x224WacDRGju6i5GIVQgtEghHR/tCxtqref1kT1o4NN+ZGjbBn7+fmE7FqIjBt7J0/1/IR58Fq0+GRg==
+"@dhis2/analytics@2.4.12":
```
martinkrulltott: We're now requesting both `@dhis2/analytics@2.4.12` (from Dashboards) and `@dhis2/analytics@2.4.9` (from the DV plugin) explicitly. Can we start using patch `~` in both places so we get `@dhis2/analytics@~2.4.12` etc so they can resolve to the same version? To avoid needing two versions of Analytics in Dashboards in the future (once DV plugin also requests a `~` patch. @janhenrikoverland @amcgee

martinkrulltott (approved):
Approved. Two minor comments that should be addressed but aren't show-stoppers.

yarn.lock
martinkrulltott: To avoid multiple Analytics versions, I suggest https://github.com/dhis2/data-visualizer-app/pull/967 is merged first along with a new version of the plugin (which uses `@dhis2/analytics@~2.4.12`). This new plugin version can then be used here along with `@dhis2/analytics@~2.4.12` to resolve both the plugin Analytics and Dashboards Analytics to `2.4.12`

src/components/Item/VisualizationItem/Item.js
jenniferarnesen: Fixed

PR #834, opened 2020-05-14
fix: translations for v32

jenniferarnesen:
Part of fix for: https://jira.dhis2.org/browse/DHIS2-8638

Things to regression test:
- interpretations
- dashboard filter
- various chart types (data-visualizer-plugin was upgraded from v32 to v33)

A sampling of fixes (filter dialog, period and org unit selector, edit mode item selector)


i18n/en.pot
```diff
@@ -100,13 +103,16 @@ msgstr ""
...
 
-msgid "See"
+msgid "See fewer {{itemTypes}}"
+msgstr ""
+
+msgid "See more {{itemTypes}}"
 msgstr ""
 
-msgid "fewer"
+msgid "+ ADD"
```
martinkrulltott, line 115: Should the + and the uppercase format be part of the translation? Or can these be handled outside of i18n? (e.g. keeping the + outside of the i18n call and transforming to uppercase using css).

martinkrulltott (approved):
LGTM. Tested locally ✅

i18n/en.pot
jenniferarnesen: Yes, I thought about that as well. I'll update it.

PR #857, opened 2020-05-20
fix: complete translations [v34] (DHIS2-8894)

jenniferarnesen:
Part of fix for: https://jira.dhis2.org/browse/DHIS2-8894

Notes:
* most changes are putting i18n.t into functions. Much of dashboards already had working translations.
* changes to test snapshots seem to be a result of regenerating the yarn.lock file (which I regenerated because I was getting unexplainable babel errors in the tests)
* period selector translations aren't yet in the analytics package (since that is still a PR and therefore hasn't been synced to transifex yet)
* I locked @dhis2/ui-core to 4.11.1 because of a css issue in the Menu component. (bug is also in master). Since I couldn't resolve the problem right away, I decided to deal with it separately rather than let it delay the translations fixes.
* **Important**, this PR depends on a new release of the DV plugin, which is in this PR:  https://github.com/dhis2/data-visualizer-app/pull/1014

Some screenshots:

Interpretations - note that some translations haven't been synced from transifex yet for the d2-ui-interpretations package. A new PR for Dashboards will be made when those are ready.


package.json
```diff
@@ -5,24 +5,23 @@
...
-        "@dhis2/d2-ui-sharing-dialog": "^6.0.1",
-        "@dhis2/d2-ui-translation-dialog": "^6.0.1",
-        "@dhis2/data-visualizer-plugin": "^34.3.34",
+        "@dhis2/d2-ui-core": "^7.0.3",
+        "@dhis2/d2-ui-interpretations": "^7.0.3",
+        "@dhis2/d2-ui-mentions-wrapper": "^7.0.3",
+        "@dhis2/d2-ui-rich-text": "^7.0.3",
+        "@dhis2/d2-ui-sharing-dialog": "^7.0.3",
+        "@dhis2/d2-ui-translation-dialog": "^7.0.3",
+        "@dhis2/data-visualizer-plugin": "^34.3.32-alpha.2",
```
janhenrikoverland: I see this is a downgrade - and also, how can I go back and check what's in this version?

martinkrulltott (approved):

package.json
jenniferarnesen: now upgraded to the correct version 34.5.2

PR #912, opened 2020-06-09
refactor: app-platform migration [TECH-339]

jenniferarnesen:
Implements https://jira.dhis2.org/browse/TECH-339

Notes:
* changes to many config files, and to package.json
* app-shell now provides HeaderBar
* prettier run on all files (vast majority of files in the PR are due to this)
* adding `toJson` results in more readable snapshots.

For reviewers:
* might be useful to step through the individual commits
* except for the following files, you can ignore all changes in `src`, because they are only prettier changes (99.9% is removing semi-colon)
* * files `AppWrapper`, `App`, `ControlBar`, `controlBarDimensions`, `DashboardVerticalOffset`, `FilterBar`, `ProgressiveLoadingContainer` have real changes and should be reviewed.
* * test and snapshot files in this commit: c307f318a179f55d127d5755fcc4ca81178bf063
* * changes related to replacing HeaderBar: 74035ab6b07638299c212937f9e12f382d37a003


.github/semantic.yml
```diff
@@ -0,0 +1,4 @@
+titleOnly: false
```
amcgee: If we want to use squash marges here this should match the one in DV https://github.com/dhis2/data-visualizer-app/blob/master/.github/semantic.yml

amcgee (changes requested):
`.env` should be removed not renamed, the local REACT_APP_DHIS2_BASE_URL env var will interfere with the platform in development

netlify.toml
```diff
@@ -13,5 +13,5 @@ ID = "c19afad0-317d-4612-a81d-950c23f62da7"
   publish = "build"
 
   # Default build command.
-  command = "yarn run build-netlify"
+  command = "yarn install && yarn build:netlify"
```
amcgee: `yarn install` should be done automatically by netlify so it shouldn't be necessary to specify here.

package.json
```diff
@@ -38,62 +38,29 @@
...
+        "enzyme-to-json": "^3.5.0",
+        "immutability-helper": "^3.1.1"
+    },
+    "jest": {
+        "setupFilesAfterEnv": [
+            "<rootDir>/src/setupTests.js"
+        ],
+        "moduleNameMapper": {
+            "^.+\\.(css|sass|scss)$": "identity-obj-proxy"
+        }
```
amcgee, line 64: This should be taken care of automatically by the platform so I don't think it's necessary here?

src/components/App.js
```diff
@@ -1,78 +1,71 @@
...
 
     getChildContext() {
-        return { baseUrl: this.props.baseUrl, i18n, d2: this.props.d2 };
+        return { baseUrl: this.props.baseUrl, i18n, d2: this.props.d2 }
     }
 
     render() {
         return (
             <>
                 <CssVariables colors spacers />
```
amcgee, line 33: Side note: we should move this to the shell

amcgee (commented):
It looks like there were a few jest snapshot files removed, why is that?

jenniferarnesen:
> It looks like there were a few jest snapshot files removed, why is that?

I don't think any were removed? Several were massively reduced in size though.

.github/semantic.yml
jenniferarnesen: Updated

netlify.toml
jenniferarnesen: Updated

package.json
jenniferarnesen: @austin When I remove it then some of the snapshot tests fail with this kind of difference: 

```   
    -       <div
    -         className="leftControls"
    -       >
    +       <div>
```

Is that what is expected, and I should update the snapshots?

jenniferarnesen:
> `.env` should be removed not renamed, the local REACT_APP_DHIS2_BASE_URL env var will interfere with the platform in development

Done.

jenniferarnesen:
This needs to get out. I'm merging after verbal approval from Jan.

PR #1022, opened 2020-08-26
fix: save default dashboard title if user does not provide a title [DHIS2-9234]

jenniferarnesen:
If there is no provided dashboard title, which is the case when starting a new dashboard, or deleting the title of the existing dashboard, a placeholder is added "Untitled dashboard" (translated). If the dashboard gets saved without the user adding a title, then the dashboard is saved with the translated "Untitled dashboard"

Placeholder:
image

Saved:
image

Multiple untitled dashboards
image


martinkrulltott (approved):
LGTM, looks pretty neat with the placeholder as well 👍

PR #1028, opened 2020-08-31
fix: correctly calculate bottom of the printable area [DHIS2-9366]

jenniferarnesen:
Calculation of the "page" bottom was incorrect.  Correct calculation code pulled into a separate function and tests added.  Make sure to check for the error from Print preview via Edit mode.

Bottom of the print preview was looking like this:

image

Fixed:  
image


janhenrikoverland (approved):
Looks good from what I can see without diving deep into it.

PR #1034, opened 2020-09-02
fix: apply scrollbar to the dashboard area below the control bars and headerbar [DHIS2-9371]

jenniferarnesen:
The overflow property of app-shell-app has to be overridden to achieve the scrollbar on the dashboard area. Then, the height of the dashboard area is calculated from window.innerHeight - any visible control/header bars.  A few margin and padding updates were needed, which affected the print "page" size.

Testing - make sure to try all dashboard modes (view, new, edit, print preview, print layout, print one-item-per-page), and change the height of the Dashboard control bar. With more time down the road after the release, this code can be cleaned up and improved.

Print mode (the top actions bar is now fixed and stays in place while scrolling):

image

View mode:
image

View mode - several control bar rows
image

Edit mode:
image

Edit mode print preview:

image


janhenrikoverland (approved):
LG for now, can be cleaned up later

PR #1038, opened 2020-09-07
fix: include title page in edit preview but scroll to below it [DHIS2-9417]

jenniferarnesen:
In contrast to the Print Layout (from View mode), the Edit print preview did not have a title page, which means there was no page break prior to the first item page. This means the first item page had more vertical space in Edit print preview than in Print Layout. The result could lead to more items getting on the first page in Edit preview, than the actual print.

The solution is to include the title page in Edit preview as well, so that Edit preview and Print Layout are identical. Then, to allow the user to focus on the important stuff (item print layout), the Edit preview is auto scrolled to the top of the first item page. (solution approved by @cooper-joe )

To test locally, create a dashboard with items the exact same height as this one:  https://play.dhis2.org/dev/dhis-web-dashboard/#/drA52w6PCfi


src/modules/printUtils.js
```diff
@@ -9,6 +9,14 @@ export const a4LandscapeWidthPx = 1102
 
 export const MAX_ITEM_GRID_HEIGHT = 33
 
+export const getTransformY = elStyle => {
+    if (!elStyle || !elStyle.transform) {
+        return null
+    }
+
+    return parseInt(elStyle.transform.split(' ')[1]?.slice(0, -3))
```
edoardo: Perhaps this won't ever happen, but what if the transform rule does not use `px` ie. `transform(5%, 50%)`?
In that case this parsing will fail.
Is this transform rule something we add so we know it will always use `px`?

edoardo (approved):
Tested locally it does what it promises 👍 

I added one comment for something that might need to be addressed.

src/modules/printUtils.js
jenniferarnesen: Even though this *should never*  happen (famous last words), I added a commit to make this code more robust

src/modules/printUtils.js
```diff
@@ -9,6 +9,22 @@ export const a4LandscapeWidthPx = 1102
 
 export const MAX_ITEM_GRID_HEIGHT = 33
 
+export const getTransformYPx = elStyle => {
+    if (!elStyle || !elStyle.transform) {
+        return null
+    }
+
+    const transformY = elStyle.transform.split(' ')[1]?.slice(0, -1)
```
edoardo: Maybe a regexp can be used here to simplify a bit as it takes care of verifying that the format is correct:
```suggestion
   const transformY = elStyle.transform.split(' ')[1].match(/(\d+)px$/)
    if (transformY) {
      return parseInt(transformY[1])
    } else {
      return null
    } 
```

edoardo (approved):

PR #1040, opened 2020-09-07
fix: scroll to top when switching dashboard

jenniferarnesen:
This functionality was broken when moving the scrollbars to the dashboard area rather than the whole document.

To reproduce the bug:
1. open a dashboard where you can scroll down
2. scroll down
3. select a different dashboard
Expected: dashboard is shown from the top
Actual: dashboard is shown with the same scrolling as the old dashboard


edoardo (approved):
The scrolling seem to happen before the dashboard is switched, so you see the current (previous) dashboard quickly scrolling all the way up before the new dashboard appears.
The scrollbar also seem to move up after the dashboard is switched, before it disappears (in Chrome OSX at least).

But it fixes the issue.

src/components/Dashboard/Dashboard.js
```diff
@@ -49,7 +49,7 @@ export class Dashboard extends Component {
 
             this.setHeaderbarVisibility()
```
amcgee, line 50: What does this do?  Shouldn't need to do anything with the headerbar in app platform...

src/components/Dashboard/Dashboard.js
```diff
@@ -49,7 +49,7 @@ export class Dashboard extends Component {
...
 
-            this.scrollToTop()
+            document.querySelector('.dashboard-wrapper')?.scroll(0, 0)
```
amcgee: To do this correctly, this should probably happen in the route handler rather than when clicking the dashboard link - that would solve the problem @edoardo saw

src/components/Dashboard/Dashboard.js
jenniferarnesen: The print view doesn't show the headerbar:

image


From Joe's sketch:

https://www.sketch.com/s/d4547123-e947-4484-89d7-bfc88e2f002f/a/dJZwDb/play

src/components/Dashboard/Dashboard.js
jenniferarnesen: Actually, moving it further down to just the View dashboard in an effect, seems to do the trick.

PR #1045, opened 2020-09-08
fix: add info in print preview about items that were shortened to fit on page [DHIS2-9423]

jenniferarnesen:
The info icon shows in print preview, but not the actual print. On hover, a tooltip appears.  The max height of an item was also increased by one unit since there is actually room on the page.

image

image


edoardo (approved):
LGTM

PR #1067, opened 2020-09-14
fix: use hooks to exert finer control on setting dashboard [DHIS2-9508]

jenniferarnesen:
Redux actions were being executed multiple times unnecessarily because of changes in props that had nothing to do with whether the dashboard needed to be set.

Implementation is a switch to using useEffect rather than class componentDidMount and componentDidUpdate.

Given a dashboard with 3 charts and 1 map:

Network tab before the fix (12 requests for charts, 4 for maps). The number of chart requests varies but it is almost always more than needed. 12 requests was the most for the 3-chart scenario:

image

Here is the network tab in dev tools after the change (6 requests for charts, 4 for maps):

image

### DIFF
--- a/src/components/Dashboard/Dashboard.js
+++ b/src/components/Dashboard/Dashboard.js
@@ -128,6 +128,7 @@ Dashboard.propTypes = {
     dashboardsIsEmpty: PropTypes.bool,
     dashboardsLoaded: PropTypes.bool,
     id: PropTypes.string,
+    match: PropTypes.object, // provided by the react-router-dom
     mode: PropTypes.string,
     routeId: PropTypes.string,
     selectDashboard: PropTypes.func,
@@ -140,7 +141,7 @@ const mapStateToProps = (state, ownProps) => {
         dashboardsIsEmpty: isEmpty(dashboards),
         dashboardsLoaded: !sDashboardsIsFetching(state),
         id: sGetSelectedId(state),
-        routeId: ownProps.match.params.dashboardId || null,
+        routeId: ownProps.match?.params?.dashboardId || null,
     }
 }
 
--- a/src/components/Dashboard/Dashboard.js
+++ b/src/components/Dashboard/Dashboard.js
@@ -71,11 +71,14 @@ export const Dashboard = ({
     }, [dashboardsLoaded, dashboardsIsEmpty, routeId, mode])
 
     useEffect(() => {
-        // TODO does this need to be cleared?
-        window.onresize = debounce(
+        const onResize = debounce(
             () => setWindowHeight(window.innerHeight),
             300
         )
+        window.addEventListener('resize', onResize)
+        return () => {
+            window.removeEventListener('resize', onResize)
+        }
     }, [])
 
     if (!dashboardsLoaded || id === null) {
--- a/src/components/Dashboard/Dashboard.js
+++ b/src/components/Dashboard/Dashboard.js
@@ -1,4 +1,4 @@
-import React, { Component } from 'react'
+import React, { useEffect } from 'react'
 import { connect } from 'react-redux'
 import PropTypes from 'prop-types'
 import isEmpty from 'lodash/isEmpty'
@@ -34,6 +34,15 @@ import {
     isPrintMode,
 } from './dashboardModes'
 
+const setHeaderbarVisibility = mode => {
+    const header = document.getElementsByTagName('header')[0]
+    if (isPrintMode(mode)) {
+        header.classList.add('hidden')
+    } else {
+        header.classList.remove('hidden')
+    }
+}
+
 const dashboardMap = {
     [VIEW]: <ViewDashboard />,
     [EDIT]: <EditDashboard />,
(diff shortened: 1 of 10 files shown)


src/components/Dashboard/Dashboard.js
```diff
@@ -42,101 +51,93 @@ const dashboardMap = {
...
         }
-    }
+    }, [dashboardsLoaded, dashboardsIsEmpty, routeId, mode])
 
-    componentDidMount() {
+    useEffect(() => {
+        // TODO does this need to be cleared?
         window.onresize = debounce(
-            () => this.props.setWindowHeight(window.innerHeight),
+            () => setWindowHeight(window.innerHeight),
```
amcgee, line 75: @jenniferarnesen yeah this effect should clean up the event handler on unmount, so something like this:

```js
useEffect(() => {
  const onResize = debounce(() => setWindowHeight(window.innerHeight))
  window.addEventListener('resize', onResize);
  return () => {
    window.removeEventListener('resize', onResize)
  }
}
```

src/components/Dashboard/Dashboard.js
```diff
@@ -42,101 +51,93 @@ const dashboardMap = {
...
+    useEffect(() => {
+        setHeaderbarVisibility(mode)
+    }, [mode])
+
+    useEffect(() => {
+        if (dashboardsLoaded && !dashboardsIsEmpty) {
+            selectDashboard(routeId)
         }
-    }
+    }, [dashboardsLoaded, dashboardsIsEmpty, routeId, mode])
```
amcgee: Is `mode` needed as a dependency here?

src/components/Dashboard/Dashboard.js
```diff
@@ -34,6 +34,15 @@ import {
...
 
+const setHeaderbarVisibility = mode => {
+    const header = document.getElementsByTagName('header')[0]
+    if (isPrintMode(mode)) {
+        header.classList.add('hidden')
+    } else {
+        header.classList.remove('hidden')
+    }
+}
+
```
amcgee, line 45: This should really be added as a feature of the platform/shell, rather than using this CSS hack.  (for another future change)

amcgee (changes requested):

src/components/Dashboard/Dashboard.js
jenniferarnesen: I can add an issue for it?

src/components/Dashboard/Dashboard.js
jenniferarnesen: Yes it is. This kind of shows how the components are a little too complex and co-dependent at the moment. But I could look to see if there is a better way.

jenniferarnesen: I'm afraid it will involve a bigger refactor of one of the thunks. I'd prefer to wait.

amcgee: What about adding the selectDashboard logic to the above effect (which has a legitimate `mode` dependency) instead?  Otherwise I worry that this might be accidentally removed in a future refactor.

src/components/Dashboard/Dashboard.js
amcgee: Thanks!  Maybe reference it in an issue on `app-platform`?

src/components/Dashboard/Dashboard.js
amcgee: The logic here with lots of inter-connected props seems like it could be refactored somehow...

src/components/Dashboard/Dashboard.js
jenniferarnesen: https://github.com/dhis2/app-platform/issues/462

src/components/Dashboard/Dashboard.js
jenniferarnesen: @amcgee Your observation led me to dig into why the `mode` prop was needed, and it led me to a TODO in my own code, which I have now fixed. So I was able to remove `mode` after all.  Some additional refactoring was needed, and I am gradually gathering all the print constants into a single place.

src/components/Dashboard/Dashboard.js
```diff
@@ -42,101 +51,96 @@ const dashboardMap = {
...
 }
 
-const mapStateToProps = state => {
+const mapStateToProps = (state, ownProps) => {
     const dashboards = sGetAllDashboards(state)
     return {
         dashboardsIsEmpty: isEmpty(dashboards),
         dashboardsLoaded: !sDashboardsIsFetching(state),
         id: sGetSelectedId(state),
+        routeId: ownProps.match.params.dashboardId || null,
```
amcgee: This will break with a reference error if `match` or `params` is undefined - we should either define an explicit shape requirement in `propTypes` or use optional chaining (`ownProps.match?.params?.dashboardId`) here

src/components/Dashboard/Dashboard.js
```diff
@@ -42,101 +51,96 @@ const dashboardMap = {
...
-        return dashboardMap[mode]
-    }
+    return dashboardMap[mode]
 }
 
 Dashboard.propTypes = {
     dashboardsIsEmpty: PropTypes.bool,
     dashboardsLoaded: PropTypes.bool,
     id: PropTypes.string,
-    match: PropTypes.object,
```
amcgee, line 128: Hmm... isn't this still needed, technically, for the final exported component?  As is there's no indication that a `match` object is required, and

src/components/Dashboard/Dashboard.js
amcgee: `match` could also be just a regular prop, and use `const routeId = match?.params?.dashboardId || null` could be at the top of the component body

src/components/Dashboard/PrintDashboard.js
```diff
@@ -63,7 +78,10 @@ export class PrintDashboard extends Component {
                 })
             }
 
-            this.props.addDashboardItem({ type: PRINT_TITLE_PAGE })
+            this.props.addDashboardItem({
+                type: PRINT_TITLE_PAGE,
+                isOipp: true,
```
amcgee: What does `Oipp` stand for?

amcgee (changes requested):
Looking good, a few questions/comments

src/components/Dashboard/Dashboard.js
jenniferarnesen: I will add that. I did move `match.params.dashboardId` from componentDidMount. It had been like that since the original dashboard code, so I figured it will always be defined. But best to be sure.

src/components/Dashboard/PrintDashboard.js
jenniferarnesen: one-item-per-page. It isn't really described anywhere though, good point.

src/components/Dashboard/Dashboard.js
jenniferarnesen: Restored

src/components/Dashboard/PrintDashboard.js
jenniferarnesen: prop renamed

src/components/Dashboard/Dashboard.js
```diff
@@ -42,101 +51,97 @@ const dashboardMap = {
...
+    match: PropTypes.object, // provided by the react-router-dom
```
amcgee, line 131: Should be required

amcgee (approved):
Code LGTM

PR #1073, opened 2020-09-17
fix: filter bar position (DHIS2-9453)

edoardo:
The position of the filter bar was wrong, too low compared to the control bar.
It was a regression after the change on the scrollbar.

Example screenshot before:
image-2020-09-09-15-33-21-046

Demo:
ezgif com-optimize (5)


jenniferarnesen (approved):

janhenrikoverland (approved):
LG

PR #1074, opened 2020-09-18
fix: dashboard filter - filter dialog incorrectly shows filter as selected even though it was removed [DHIS2-9560]

jenniferarnesen:
The FilterDialog component was always rendered, because FilterSelector is always rendered in the ViewTitleBar. That meant that the `filter` state was never reset, causing previous item filter to still show.

Fix includes:
* only render the FilterDialog if there is a selected dimension
* move item handling (onSelect, onDeselect...) to the FilterDialog. Since FilterDialog is removed when closed, it will be able to pick up the current state from redux in the `useState` statement when reopened.


src/components/ItemFilter/FilterDialog.js
```diff
@@ -20,21 +18,73 @@ import {
...
+        setFilters({ [dimensionId]: items })
+    }
 
+    const onDeselectItems = ({ dimensionId, itemIdsToRemove }) => {
+        const oldList = filters[dimensionId] || []
+        const newList = oldList.filter(
+            item => !itemIdsToRemove.includes(item.id)
+        )
+
+        const list = newList.length ? newList : []
```
edoardo: Is this necessary?
If newList is empty, it's already `[]`

edoardo (approved):

src/components/ItemFilter/FilterDialog.js
jenniferarnesen: Not sure. I just copied this code over from FilterSelector. Will check.

jenniferarnesen: Removed the unneeded line

PR #1075, opened 2020-09-18
fix: reduce renders caused by state change

jenniferarnesen:
orObject and orArray cause new objects to be created, and this could cause a component to think something has changed when it may not have.

Also, FilterDialog was rerendering several times even because of the `selectedDimensions` property. Turns out this property isn't really needed. By removing that property, the number of renders was reduced to 1.

I doubt this alone would have much if any performance effect, but all the little stuff does add up.

### DIFF
--- a/src/actions/selected.js
+++ b/src/actions/selected.js
@@ -1,4 +1,8 @@
-import { getCustomDashboards, sGetDashboardById } from '../reducers/dashboards'
+import {
+    getCustomDashboards,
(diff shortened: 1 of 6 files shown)


jenniferarnesen:
This one has been merged into https://github.com/dhis2/dashboards-app/pull/1074

jenniferarnesen:
Merge with a different PR

PR #1083, opened 2020-09-21
fix: unmount map when switching dashboard mode or changing active type [DHIS2-9558]

jenniferarnesen:
When switching modes, the components are unmounted and mounted anew. But `plugin.unmount` was not being called on maps plugin, resulting in the warning about too many WebGL contexts.

Also removed the call to `plugin.unmount` in the onSelectActiveType handler since the component gets unmounted anyway (ComponentWillUnmount) which will call the `plugin.unmount`


janhenrikoverland (approved):
LGTM

PR #1084, opened 2020-09-21
fix: set passive to true for scroll event listener in ProgressiveLoadingContainer [DHIS2-9508]

jenniferarnesen:
A minor optimization that probably won't have much impact on the performance, but nevertheless doesn't have a negative impact.

Changes include:
- do not add eventListener on the `window`
- set `passive: true` on the event listener (see explanation here: https://web.dev/uses-passive-event-listeners/?utm_source=lighthouse&utm_medium=devtools)


amcgee (approved):
Code looks good to me - we're sure the `window` handler is unnecessary?  Probably is...

jenniferarnesen:
> Code looks good to me - we're sure the `window` handler is unnecessary? Probably is...

I tested the dashboard scroll in Chrome, FF, and Safari. When I get access, will also test with Edge. So far the scrolling works.

Can you think of a reason why we would need the event listener on `window`?

amcgee:
No, I think I just added it as a fallback - in full-screen layouts we shouldn't have any window scroll events

PR #1104, opened 2020-09-23
fix: if editing then use the original vis type [DHIS2-9589]

jenniferarnesen:
When editing a dashboard, the original visualization type should be displayed. The user may have changed it in View mode (i.e., chart -> table).  In the edit print preview, the original vis type should display, while when printing from View mode, the active vis type should display.

Small refactor:
- renamed a couple of actions to better fit what it does and used the simpler mapDispatchToProps form.


jenniferarnesen:
This is part of another PR

PR #1111, opened 2020-09-24
fix: changing View As on one item was incorrectly causing other items to also change view [DHIS2-9590]

jenniferarnesen:
Fixes:

* if 2 dashboard items have the same visualization, switching the "View as" on one item should not affect the other item
* if 2 dashboards have the same visualization, and the "View as" has been switched on one dashboard, it should not affect the other dashboard
* also fixes [DHIS2-9589]: Edit mode should show the original visualization type, not the "view as" type.

Fix for 9590 involves separating the "Active type" from the "Visualization" in redux, since the visualization is in a map keyed by visualization id (which can be shared between dashboard items). Fix also includes clearing out some of the redux properties when switching dashboards (clear visualizations)

Fix for 9589 was to choose the vis type to display based on whether the user is currently editing. In edit mode (and edit print preview), it should show the original visualization


martinkrulltott (approved):
Not previously familiar with this part of dashboards but LGTM. Tested locally, the fix is working as expected, however I'm having issues with maps dashboard items that aren't loading (see [Slack discussion](https://dhis2.slack.com/archives/G0108CD1421/p1600931854000200)). I assume this is unrelated to this fix though.

PR #1149, opened 2020-10-07
fix: use ui alertbar instead of material-ui

jenniferarnesen:
The following changes have been made in this PR:
* switch from material-ui to @dhis2/ui AlertStack/AlertBar.
** One consequence of this is that we lose the formatting (bold) of the dashboard name in the message.
** Slight visual difference - previously the bar rested on the bottom of the viewport, now it is positioned slightly above the bottom
* rename snackbar to alert for reducers, actions and the component
* simplify the alert redux store to only include the message string (the original code was potentially overengineered)

Before, with material-ui
image

Aftr, with @dhis2/ui AlertBar
image


src/actions/selected.js
```diff
@@ -75,18 +70,12 @@ export const acClearSelectedItemActiveTypes = () => ({
...
+    const alertTimeout = setTimeout(() => {
+        const name = sGetDashboardById(getState(), id)?.displayName
 
+        if (sGetSelectedIsLoading(getState()) && name) {
             dispatch(
-                acReceivedSnackbarMessage({
-                    message: loadingDashboardMsg,
-                    open: true,
-                })
+                acSetAlertMessage(`${i18n.t('Loading dashboard')}: ${name}`)
```
edoardo: The string should probably have a placeholder for the name.
Some languages might need to position the name before the rest of the text, also, the `: ` can cause problems (ie. RTL)

src/components/AlertBar/AlertBar.js
```diff
@@ -0,0 +1,29 @@
...
+import { connect } from 'react-redux'
+import { AlertBar, AlertStack } from '@dhis2/ui'
+import PropTypes from 'prop-types'
+
+import { sGetAlertMessage } from '../../reducers/alert'
+import { acClearAlertMessage } from '../../actions/alert'
+
+export const Alert = ({ message, onClose }) =>
+    message ? (
+        <AlertStack>
```
edoardo, line 11: Not sure we need `AlertStack` if we only have 1 message at a time...

src/components/AlertBar/AlertBar.js
```diff
@@ -0,0 +1,29 @@
...
+            <AlertBar onHidden={onClose} permanent>
```
edoardo, line 12: For DV I used the `variant` prop for different types of alerts, and the `permanent` prop is only set for errors or warnings.
Ie. you would want the loading message to disappear automatically. That is built in so you don't have to handle the timeout in the app.
See: https://github.com/dhis2/data-visualizer-app/blob/master/packages/app/src/components/Snackbar/Snackbar.js#L16-L37

cooper-joe:
Alert bar usage LGTM 👍

cooper-joe (approved):

src/actions/selected.js
jenniferarnesen: Then we should find another separator. I can't put ":" into the translation string bc it is used as a namespace separator

src/components/AlertBar/AlertBar.js
jenniferarnesen: The AlertStack provides the positioning, so we need it.

src/actions/selected.js
edoardo: True. But regardless of the `:` problem, the interpolation with the placeholder still applies.

src/components/AlertBar/AlertBar.js
edoardo: ok, i should probably update DV then.

src/components/AlertBar/AlertBar.js
jenniferarnesen: The way the alertbar currently works is that the message stays until  the app detects that the dashboard is finished loading. So I actually want the message to be permanent in this case. When the loading is complete, then the message in redux is cleared and the Alert renders null.

src/actions/selected.js
jenniferarnesen: @edoardo updated to use interpolation

src/components/AlertBar/AlertBar.js
edoardo: ok, sounds good! 👍

src/actions/selected.js
```diff
@@ -75,18 +70,14 @@ export const acClearSelectedItemActiveTypes = () => ({
...
+        const name = sGetDashboardById(getState(), id)?.displayName
 
+        if (sGetSelectedIsLoading(getState()) && name) {
             dispatch(
-                acReceivedSnackbarMessage({
-                    message: loadingDashboardMsg,
-                    open: true,
-                })
+                acSetAlertMessage(
+                    i18n.t(`Loading dashboard – ${name}`, { name })
```
edoardo: 🤔  this doesn't look like is using the correct syntax:
https://www.i18next.com/translation-function/interpolation

I suspect is not doing what we want?
```suggestion
                    i18n.t("Loading dashboard – {{name}}", { name })
```

jenniferarnesen: Oops, you're right 😅

edoardo (approved):

PR #1183, opened 2020-10-19
fix: ou item filter on App items was crashing Dashboards [DHIS2-9725]

jenniferarnesen:
The critical change is a one-liner here: https://github.com/dhis2/dashboards-app/pull/1183/files#diff-b2b6d6eb213be7f4762358a9fd09214140524af90574d6ea65f90deae961e637R18
( `ouPath`  --> `ouFilter.path`)

In addition, some cleanup was done:
- create a redux selector to get the ou item filters rather than directly accessing the state
- added unit test for the AppItem

### DIFF
--- a/src/components/Item/AppItem/Item.js
+++ b/src/components/Item/AppItem/Item.js
@@ -3,26 +3,19 @@ import PropTypes from 'prop-types'
 import { connect } from 'react-redux'
 import NotInterestedIcon from '@material-ui/icons/NotInterested'
 
-import { FILTER_ORG_UNIT } from '../../../actions/itemFilters'
-import {
-    sGetItemFiltersRoot,
-    DEFAULT_STATE_ITEM_FILTERS,
-} from '../../../reducers/itemFilters'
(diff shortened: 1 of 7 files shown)


src/components/Item/AppItem/Item.js
```diff
@@ -31,7 +24,7 @@ const getIframeSrc = (appDetails, item, itemFilters) => {
     return iframeSrc
 }
 
-const AppItem = ({ dashboardMode, item, itemFilters }, context) => {
+const AppItem = ({ dashboardMode, item, ouItemFilters }, context) => {
```
edoardo: The fix looks good.
It's been a while, but I think the original idea here was to pass all the filters in `itemFilters` and pass them to the iframe URL.
Apparently only the ou one was implemented, but I'm wondering if we should keep `itemFilters` for when other filters need to be passed in (like period).
Perhaps all filters should already be passed?
Probably not all AppItem widgets support all of them but wouldn't it be more correct to pass all (not sure about the URL query parameters naming)?

edoardo (approved):

PR #1207, opened 2020-10-22
feat: add support for updating favorite data statistics [DHIS2-9554]

tomzemp:
To implement functionality to count data statistics for favorite items viewed from within a dashboard, relevant actions/dispatches have been added and reducers expanded to access/persist system setting keyGatherAnalyticalObjectStatisticsInDashboardViews.

VisualizationItem/Item has been amended to check keyGatherAnalyticalObjectStatisticsInDashboardViews and initiate a post to api/datastatistics for favorite. Relevant additional code modified in VisualizationItem/plugin, api/dataStatistics, and modules/itemTypes


tomzemp:
Two notes:
- This feature is being requested by PEPFAR (hence why I am submitting this pull request)
- I have where possible reused code from https://github.com/dhis2/data-visualizer-app to implement this feature (to aim for general alignment with your coding approach/style and because this seems consistent with the existing similarities between these repositories' code). 
&nbsp;
In doing this, I have tried to leave as much unchanged as possible, so for example, leaving comments untouched (https://github.com/tomzemp/dashboards-app/blob/master/src/api/settings.js#L15-L16 vs https://github.com/dhis2/data-visualizer-app/blob/master/packages/app/src/api/settings.js#L15-L16). I assumed it would be better to let you decide whether you prefer consistency or cleaning up/paring down the code more before reuse here in the dashboards app.

Thanks!

src/modules/settings.js
```diff
@@ -0,0 +1,17 @@
+export const SYSTEM_SETTINGS = [
+    //'keyDateFormat',
+    //'keyAnalysisRelativePeriod',
+    //'keyAnalysisDigitGroupSeparator',
+    //'keyIgnoreAnalyticsApprovalYearThreshold',
+    'keyGatherAnalyticalObjectStatisticsInDashboardViews',
+];
+
+export const extractUserSettings = settings => {
+    const nameProp = settings.keyAnalysisDisplayProperty;
```
janhenrikoverland: This function does not seem to be used for anything.

src/components/Item/VisualizationItem/Item.js
```diff
@@ -69,6 +70,14 @@ export class Item extends Component {
             await pluginManager.fetch(this.props.item)
         )
 
+        try {
+            if (this.props.countInDashboard) {
+                const postFavoriteDataStatistics = await pluginManager.postFavoriteDataStatistics(this.props.item)
```
jenniferarnesen: pluginManager has been refactored so that it only contains functions relevant to the maps, event reports, event visualizer plugins.

So instead of adding this function to the pluginManager, just call the api directly here.  (and use function `getVisualizationId` from modules/item)

src/components/Item/VisualizationItem/Item.js
```diff
@@ -328,6 +337,7 @@ Item.contextTypes = {
 
 Item.propTypes = {
     activeType: PropTypes.string,
+    countInDashboard: PropTypes.bool,
```
jenniferarnesen: What about naming this "gatherDataStatistics"?

janhenrikoverland:
@tomzemp Thanks for the PR. We have a few comments, other than that it looks fine!

src/actions/settings.js
```diff
@@ -0,0 +1,38 @@
...
+    }
+
+    const onError = error => {
+        console.log('Error (apiFetchSystemSettings): ', error)
+        return error
+    }
+
+    try {
+        const systemSettings = await apiFetchSystemSettings();
+        //const rootOrganisationUnit = await apiFetchOrganisationUnitRoot()
```
jenniferarnesen: remove commented code

src/api/settings.js
```diff
@@ -0,0 +1,22 @@
...
+export const apiFetchSystemSettings = () => {
+    const endPoint = '/systemSettings'
+    const url = `${endPoint}?${SYSTEM_SETTINGS.map(s => `key=${s}`).join('&')}`
+
+    return getInstance()
+        .then(d2 => d2.Api.getApi().get(url))
+        .catch(onError)
+}
+
+export const apiFetchUserSettings = () => {
```
jenniferarnesen: This function doesn't seem to be used?

src/api/settings.js
```diff
@@ -0,0 +1,22 @@
+import { getInstance } from 'd2'
+import { onError } from './index'
+import { SYSTEM_SETTINGS } from '../modules/settings'
```
jenniferarnesen: If Jan is correct in his comment below that `extractUserSettings` isn't used, then I suggest just getting rid of `modules/settings` and use `keyGatherAnalyticalObjectStatisticsInDashboardViews` directly in this file. If more settings are needed in the future, then we can expand this.

src/modules/settings.js
tomzemp: Agreed. Left over from importing from data-visualizer-app. Removing along with rest of modules/settings.

src/components/Item/VisualizationItem/Item.js
tomzemp: Got it. Updated this.

src/components/Item/VisualizationItem/Item.js
tomzemp: Sounds good. Renamed.

src/actions/settings.js
tomzemp: Removed.

src/api/settings.js
tomzemp: Agreed. Just left over from importing from data-visualizer-app. Removing.

src/api/settings.js
tomzemp: Sounds good. Removed modules/settings.

tomzemp:
Thanks for the review! I've gone and cleaned up the code to incorporate your comments. I've removed unused lines from the code imported from data-visualizer-app, moved the call to api/dataStatistics to be within VisualizationItem/Item (and hence reverted changes I had for VisualizationItem/plugin), removed modules/settings, and changed the property name in VisualizationItem/Item.

I've just made incremental changes, which work when I merge back into the latest copy of the repo and pick up the refactoring that you've done, but I could also submit a new pull request starting from the current app. Let me know if there's anything else I can do. Thanks!

jenniferarnesen (approved):

jenniferarnesen:
@tomzemp

With the implementation in the PR, statistics will be posted every time the user changes the view while in dashboard. ie, when they:
1: open the app
2) go into edit mode,
3) go into edit print preview,
4) exit edit print preview,
5) exit edit mode,
6) enter print layout,
7) exit print layout,
8: enter print one-item-per-page,
9) exit print one-item-per-page
10) change dashboard (similar to #1)
Just want to confirm that is the desired behavior.

According to Jan, only 1, 5, and 10 should trigger a postDataStatistics. That means we need to figure out how to only call it at those points.
1) when navigating to a dashboard url (e.g., following a dashboard link, when user manually types in a url).
2) When user clicks on a dashboard chip
3) When user clicks either "Save changes" or "Exit without saving".

For 1 and 2, I think it will work if you make the api request here: https://github.com/dhis2/dashboard-app/blob/master/src/actions/selected.js#L107
For 3, maybe it's possible to make the api request when the user clicks on either of those buttons.

src/api/dataStatistics.js
```diff
@@ -1,8 +1,20 @@
...
 
 export const apiPostDataStatistics = async (eventType, id) => {
     const d2 = await getInstance()
     const url = `dataStatistics?eventType=${eventType}&favorite=${id}`
 
     d2.Api.getApi().post(url)
 }
+
+// update data statistics
+export const apiPostFavoriteDataStatistics = async (id, type) => {
```
jenniferarnesen: This is almost exactly the same code as the function right above it. It should be possible to just use the existing function, and instead get the dataStatisticsName at the point the function is called (currently in Item.js)

tomzemp:
Thanks @jenniferarnesen re: https://github.com/dhis2/dashboard-app/pull/1207#issuecomment-737852651

I've spoken with Tim Harding at PEPFAR, and he thinks Jan's suggestion is good, but is also pretty open to anything that results in these items being included in dashboardstatistics even if it errs on the side of over counting :-)

With that in mind, I was thinking that maybe it would be more feasible to count when an item is rendered in view mode (https://github.com/tomzemp/dashboard-app/commit/23df1978c4eba910329d86fb4f1054a66503e161#diff-377f44d230e754f6632e36e95b2063eee06910961fcabeaf54d2936f14503b11R76). This would meet 1,5,10 as I understand, but would also pick up 9 and potentially 7 (picks up if you open 'Print: Dashboard Layout' from view mode and then exit back to view mode, but does not pick up if you open 'Print preview' from within Edit mode and then exit back to edit mode)

If that's acceptable on your end, it's acceptable for PEPFAR. If you want the logic to be more refined and to only pick up 1,5,10, let me know and I can go back and make other changes. If I'm missing something, let me know. Thanks!

src/api/dataStatistics.js
tomzemp: That would be better 😄 

I've refactored: 
https://github.com/tomzemp/dashboard-app/commit/1223f80404e0504cd7d086cfb5b649e5a1939d99

jenniferarnesen:
> Thanks @jenniferarnesen re: [#1207 (comment)](https://github.com/dhis2/dashboard-app/pull/1207#issuecomment-737852651)
> 
> I've spoken with Tim Harding at PEPFAR, and he thinks Jan's suggestion is good, but is also pretty open to anything that results in these items being included in dashboardstatistics even if it errs on the side of over counting :-)
> 
> With that in mind, I was thinking that maybe it would be more feasible to count when an item is rendered in view mode ([tomzemp@23df197#diff-377f44d230e754f6632e36e95b2063eee06910961fcabeaf54d2936f14503b11R76](https://github.com/tomzemp/dashboard-app/commit/23df1978c4eba910329d86fb4f1054a66503e161#diff-377f44d230e754f6632e36e95b2063eee06910961fcabeaf54d2936f14503b11R76)). This would meet 1,5,10 as I understand, but would also pick up 9 and potentially 7 (picks up if you open 'Print: Dashboard Layout' from view mode and then exit back to view mode, but does not pick up if you open 'Print preview' from within Edit mode and then exit back to edit mode)
> 
> If that's acceptable on your end, it's acceptable for PEPFAR. If you want the logic to be more refined and to only pick up 1,5,10, let me know and I can go back and make other changes. If I'm missing something, let me know. Thanks!

What it measures isn't up to me as a developer. It is for the user. So I guess it should be agreed between Tim and Scott (Russpatrick), as this implementation will affect all programs that use it.

benguaraldi:
> What it measures isn't up to me as a developer. It is for the user. So I guess it should be agreed between Tim and Scott (Russpatrick), as this implementation will affect all programs that use it.

@jenniferarnesen I'm not sure we need to go back and review this with Scott, as this feature has already been reviewed by Lars, Phil, Jim Grace, and others.

Perhaps the most straightforward route would be to implement it as-is (as this is a big improvement for us) and then if anyone uses it and requests changes, do that then?

jenniferarnesen:
Merged from a separate PR (so the verify steps get run)

PR #1228, opened 2020-10-27
test: add cypress config and tests [TECH-464]

jenniferarnesen:
Included in this PR:
* cypress configuration based on current DHIS2 convention (cli-utils-cypress initialized the configuration)
* tests for viewing dashboard (can be run against actual DHIS2 instance or stubbed fixtures - useful for CI)
* tests for creating and deleting dashboard (can only be run against an actual DHIS2 instance)
* fixtures for non-mutating tests (viewing dashboard)
* github action to run non-mutating tests on CI when creating and updating PRs, and on push to master
* README updated with info on how to set up and run the tests
* linting forced a function name change for a deprecated react lifecycle method (componentWillReceiveProps to UNSAFE_componentWillReceiveProps). Doesn't affect functionality.

This PR in the first step in building e2e test coverage. I expect that the tests will get reorganized into different files/folders as the test suite grows.

You can see the test run on the [Cypress dashboard for dashboards-app](https://dashboard.cypress.io/projects/5fk191/runs?branches=%5B%5D&committers=%5B%5D&flaky=%5B%5D&page=1&status=%5B%5D&tags=%5B%5D&timeRange=%7B%22startDate%22%3A%221970-01-01%22%2C%22endDate%22%3A%222038-01-19%22%7D)


cypress/integration/ui/view_dashboard/index.js
```diff
@@ -0,0 +1,93 @@
...
+
+    cy.get('[data-test="dhis2-dashboard-view-dashboard-title"]')
+        .should('be.visible')
+        .and('contain', 'Antenatal Care')
+    cy.get('.highcharts-background').should('exist')
+})
+
+When('I select the Immunization dashboard', () => {
+    cy.get('[data-test="dhis2-uicore-chip"]')
+        .contains('Immun')
```
martinkrulltott, line 26: Note that other dashboard items starting with "immun" will also be targeted here, causing the test to crash or behave incorrectly. I would target the whole name (still not 100% safe if someone adds another item e.g. "some name NEW") or add the id of the dashboard item to the item itself, like `data-test={`dhis2-uicore-chip-${dashboardId}`} to be sure you always get the right item

cypress/integration/ui/view_dashboard/index.js
```diff
@@ -0,0 +1,93 @@
+import { Given, When, Then } from 'cypress-cucumber-preprocessor/steps'
+
+const antenatalCareDashboardRoute = '#/nghVC4wtyzi'
+const immunizationDashboardRoute = '#/TAMlzYkstb7'
+
+Given('I open the Antenatal Care dashboard', () => {
```
martinkrulltott, line 6: Would it be possible to keep `Antenatal Care` in a variable to easily switch it out if the DB changes or we decide its better to use another dashboard as our test dashboard? Or to pick it out of an array to loop through this multiple times for different dashboards?

cypress/integration/ui/view_dashboard/index.js
```diff
@@ -0,0 +1,93 @@
...
+        .contains('Antenatal Care')
+        .click()
+})
+
+Then('the Antenatal Care dashboard displays in view mode', () => {
+    cy.location().should(loc => {
+        expect(loc.hash).to.equal(antenatalCareDashboardRoute)
+    })
+
+    cy.get('[data-test="dhis2-dashboard-view-dashboard-title"]')
```
martinkrulltott, line 18: This is currently up for discussion in our coming frontend meeting, but I would argue for using something like `getBySel` ([source](https://docs.cypress.io/guides/references/best-practices.html#Real-World-Example)) to shorthand the `data-test` and/or the `dhis2-dashboard` prefix to avoid repetition.

cypress/integration/ui/view_dashboard/index.js
```diff
@@ -0,0 +1,93 @@
...
+        .and('contain', 'Immunization')
+    cy.get('.highcharts-background').should('exist')
+})
+
+When('I search for dashboards containing Immunization', () => {
+    cy.get('[data-test="dhis2-dashboard-search-dashboard-input"]').type('Immun')
+})
+
+Then('Immunization and Immunization data dashboards are choices', () => {
+    cy.get('[data-test="dhis2-uicore-chip"]')
```
martinkrulltott, line 46: As mentioned above I would argue its better to append the id of the item that should be visible and target that specifically, which truly will make sure that 'Immunization and Immunization data dashboards are choices', instead of the current implementation which only checks that "two dashboards are choices"

martinkrulltott: Also `.and('have.length', 2)` means adding another dashboard containing `Immun` will cause this to fail.

cypress/integration/ui/view_dashboard/index.js
```diff
@@ -0,0 +1,93 @@
...
+When('I click to preview the print one-item-per-page', () => {
+    cy.get('[data-test="dhis2-dashboard-more-button"]').click()
+    cy.get('[data-test="dhis2-dashboard-print-menu-item"]').click()
+    cy.get('[data-test="dhis2-dashboard-print-oipp-menu-item"]').click()
+})
+
+Then('the print one-item-per-page displays', () => {
+    //check the url
+    cy.location().should(loc => {
+        expect(loc.hash).to.equal(`${antenatalCareDashboardRoute}/printoipp`)
```
martinkrulltott, line 88: In favour of DRY I would abstract `location().should(loc => { expect(loc.hash).to.equal`) }` to a helper function (e.g. `expectLocationToBe(url)`, but I know some people would disagree with this approach 🤷‍♂️

cypress/integration/ui/create_dashboard/index.js
```diff
@@ -0,0 +1,111 @@
...
+    ROUTE_PRINTLAYOUT,
+    ROUTE_PRINTOIPP,
+]
+
+const getRouteFromHash = hash => {
+    const lastSlashIdx = hash.lastIndexOf('/')
+    return hash.slice(lastSlashIdx + 1)
+}
+
+const toggleShowMoreButton = () => {
```
martinkrulltott, line 26: Extract this to a separate file containing all the helper functions for the toolbar buttons?

cypress/integration/ui/create_dashboard/index.js
```diff
@@ -0,0 +1,111 @@
...
+When('dashboard title is added', () => {
+    cy.get('[data-test="dhis2-dashboard-dashboard-title-input"]').type(
+        dashboardTitle
+    )
+})
+
+When('dashboard items are added', () => {
+    cy.get('[data-test="dhis2-dashboard-item-search"]').click()
+    cy.get(
+        '[data-test="dhis2-dashboard-menu-item-ANC: 1 and 3 coverage Yearly"]'
```
martinkrulltott, line 47: Use the id instead of the name? `[data-test="dhis2-dashboard-menu-item-ANC: 1 and 3 coverage Yearly"]` looks a bit funny 🙂

cypress/integration/ui/create_dashboard/index.js
```diff
@@ -0,0 +1,111 @@
...
+    ).click()
+    // close the item selector
+    cy.get('body').trigger('keydown', { key: 'Escape' })
```
martinkrulltott, line 52: Extract this to a separate step e.g. "and the item selector is closed" that also asserts that the item selector is no longer visible?

cypress/integration/ui/create_dashboard/index.js
```diff
@@ -0,0 +1,111 @@
+import { Given, When, Then } from 'cypress-cucumber-preprocessor/steps'
+
+const UID_LENGTH = 11
+
+const dashboardTitle = new Date().toString()
```
martinkrulltott: Consider changing this to something like `TEST_DASHBOARD_TITLE` to indicate that it's a test string?

cypress/integration/ui/create_dashboard/index.js
```diff
@@ -0,0 +1,111 @@
...
+        dashboardTitle
+    )
+})
+
+Then('dashboard displays in view mode', () => {
+    cy.location().should(loc => {
+        const currentRoute = getRouteFromHash(loc.hash)
+
+        expect(nonViewRoutes).not.to.include(currentRoute)
+        expect([0, UID_LENGTH]).to.include(currentRoute.length)
```
martinkrulltott: So the route length can be either 0 or 11? Maybe include a comment explaining why or what this does?

martinkrulltott (approved):
LGTM. The focus of this PR isn't on the tests themselves, so I've included my comments only as suggestions for future improvements, rather than change requests. 
I'm not too familiar with the Cucumber syntax, but generally I think the steps should be named in a way that they only contain what the name implies. e.g. 
` Then('the dashboard displays in edit mode', () => {` 
    ` cy.get('[data-test="dhis2-dashboard-dashboard-title-input"]').should(
        'exist'
    )` 
    ` cy.location().should(loc => {
        expect(getRouteFromHash(loc.hash)).to.eq(ROUTE_EDIT)
    })` 
` })`
are really a two-in-one, that both checks that the item displays and that the url is correct. Preferably I'd split that in to separate steps (or rename the one step to make it more apparent that two things are checked). But this could also caused by my lack of experience in Cucumber... 🤷‍♂️ 
____ 
Anyway, looks like we're off to a good start for a full test suite 👍 🥇

cypress/integration/ui/create_dashboard/index.js
martinkrulltott: I think this is a pretty clever way to name the items created by the tests. You can use `toUTCString` or `toLocaleString` to get a more compact version (without the `GMT+0100 (Central European Standard Time)` part)

cypress/integration/ui/view_dashboard/index.js
jenniferarnesen: Actually, I did this on purpose. "Immun" will have 2 results, and hitting the enter key should choose the first one.

cypress/integration/ui/create_dashboard/index.js
jenniferarnesen: Done

jenniferarnesen: Done

cypress/integration/ui/create_dashboard/index.js
jenniferarnesen: Done

cypress/integration/ui/create_dashboard/index.js
jenniferarnesen: The id isn't uniformly available for all items. I'll revisit this when I add a new set of tests

cypress/integration/ui/create_dashboard/index.js
jenniferarnesen: I was waiting to do this kind of things until I had more tests and see what kind of helper files make sense

cypress/integration/ui/view_dashboard/index.js
jenniferarnesen: Will update my tests according to the decision made in the upcoming meeting

cypress/integration/ui/view_dashboard/index.js
jenniferarnesen: https://github.com/dhis2/dashboards-app/blob/master/src/components/ControlBar/DashboardsBar.js#L77

jenniferarnesen:
> I'm not too familiar with the Cucumber syntax, but generally I think the steps should be named in a way that they only contain what the name implies. e.g.
> ` Then('the dashboard displays in edit mode', () => {`
> ` cy.get('[data-test="dhis2-dashboard-dashboard-title-input"]').should( 'exist' )`
> ` cy.location().should(loc => { expect(getRouteFromHash(loc.hash)).to.eq(ROUTE_EDIT) })`
> ` })`
> are really a two-in-one, that both checks that the item displays and that the url is correct. Preferably I'd split that in to separate steps (or rename the one step to make it more apparent that two things are checked). But this could also caused by my lack of experience in Cucumber... 🤷‍♂️

I guess I look at it based on this "best practice":  https://docs.cypress.io/guides/references/best-practices.html#Creating-%E2%80%9Ctiny%E2%80%9D-tests-with-a-single-assertion

Suppose a tester or product person is writing the test feature file (which is the idea, though it probably never will happen). They would not break down each assertion (url is correct, title is correct, the first item's vis is correct) into separate clauses in the Gherkin sentence. They just want it checked that "the right page with the right information is showing".

That said, I am by no means a Gherkin expert and maybe the sentence clauses should be written differently. Maybe that is a good topic for a test-automation meeting.

cypress/integration/ui/create_dashboard/index.js
jenniferarnesen: I added a constant ROOT_ROUTE_LENGTH=0 with a comment

cypress/integration/ui/view_dashboard/index.js
jenniferarnesen: Will revisit this when I add the new set of tests

PR #1265, opened 2020-11-04
chore: automerge all minor dependency PRs

janhenrikoverland:
Would like to wait a bit longer and see how the [kodiak](https://github.com/dhis2/notes/issues/71#issuecomment-716535285) experiments are going. In the meantime I want dependabot to automerge all minor deps.

### DIFF
--- a/.dependabot/config.yml
+++ b/.dependabot/config.yml
@@ -8,9 +8,5 @@ update_configs:
       target_branch: 'master'
       automerged_updates:
           - match:
-                dependency_name: '@dhis2/*'
                 dependency_type: 'all'
                 update_type: 'semver:minor'
-          - match:
-                dependency_type: 'all'
-                update_type: 'security:patch'


amcgee (commented):
The code change looks fine, but there is some risk in auto-merging even 3rd party deps.  Will defer to @varl as he's more familiar with the auto-merge strategy

janhenrikoverland:
@amcgee I based it on @varl 's findings:

"My experience (so far) based on validating and merging hundreds of these dependency upgrades over the course of a year is that when quickly merging patch and minor releases they rarely pose a problem"

janhenrikoverland:
@amcgee @varl  We've also had this setting in DV for about a year.

varl:
At the end of the day if we automatically merge dependency updates out of convenience for us, we are going to have something break _eventually_. The question is if we have good enough mechanisms to catch those breakages early and fix them quickly.

I recommend changing the `update_schedule` to 'weekly' or 'monthly':

```
update_schedule: 'daily'
```

varl (commented):
See above comment for recommendation about `update_schedule`, other than that I think it's OK.

varl (approved):

janhenrikoverland:
@varl Thanks and I see your point about `update_schedule`. However part of the strategy is to let dependabot take care of our own @dhis2 dependencies as well, e.g. there's a fix in the DV plugin and we want to propagate it to the dashboard immediately (i.e. same day) without doing that work manually. Is there a way to e.g. set @dhis2 dependencies to `daily` and everything else to `monthly`?

varl:
Not to my knowledge. Ultimately it comes down choosing a strategy that the team is comfortable living with and it sounds like you have that.

In the future we will have another way to quickly propagate dhis2-packages, so then the update schedule can be reconfigured without losing that benefit.

janhenrikoverland:
@varl Good reflections

PR #1304, opened 2020-11-17
refactor: replace material-ui popover with ui popover in Item menu and ItemSelector

jenniferarnesen:
ItemSelector refactor:
* switch to functional component with hooks
* put ItemSearchField into separate file

Item menu (notice that the popover now appears below the button instead of over it).

before: 
image

after:
image

ItemSelector:

before:
image

after:
image


cooper-joe (approved):
Functionally looks good. 

By the way, do you know anything about a fix for the "double border" appearing at the bottom? I cannot recall if this was being fixed in `ui`, or if it is a `material-ui` and `ui` clashing issue.

edoardo:
> Functionally looks good.
> 
> By the way, do you know anything about a fix for the "double border" appearing at the bottom? I cannot recall if this was being fixed in `ui`, or if it is a `material-ui` and `ui` clashing issue.

I can answer to that, it's a `ui` issue.
I reported it [here](https://jira.dhis2.org/browse/LIBS-69).

src/components/Item/VisualizationItem/ItemHeaderButtons.js
```diff
@@ -105,18 +111,23 @@ const ItemHeaderButtons = (props, context) => {
...
+    const buttonRef = createRef()
+
     return pluginIsAvailable(activeType || item.type) ? (
         <>
-            <Button small secondary onClick={handleMenuClick}>
-                <ThreeDots />
-            </Button>
-            {anchorEl && (
+            <div ref={buttonRef}>
+                <Button small secondary onClick={handleMenuClick}>
```
edoardo: @cooper-joe since now the button is always visible, should it toggle the menu?

src/components/Item/VisualizationItem/ItemHeaderButtons.js
```diff
@@ -54,16 +60,16 @@ const ItemHeaderButtons = (props, context) => {
 
     const itemHasMapView = () => hasMapView(item.type)
 
-    const handleMenuClick = (_, event) => setAnchorEl(event.currentTarget)
+    const handleMenuClick = () => setMenuIsOpen(true)
```
edoardo: I'm being picky here, but perhaps rename this to `handleButtonClick` or similar, it's not a click on the Menu.
Or if this becomes a toggle, `toggleMenu` could work.

edoardo (approved):
Looks good.

src/components/Item/VisualizationItem/ItemHeaderButtons.js
edoardo: actually it might already do that because of the `onClickOutside` prop in `Popover`...

PR #1314, opened 2020-11-19
test: add cypress test coverage

jenniferarnesen:
Run CI against a dedicated dhis2 instance. This is because there is currently a problem generating stubs for all the requests. For the tests that switch a dashboard item from e.g. chart -> table, some of the analytics requests for the table are not being captured and I don't yet know why.

Reorganizing the code


jenniferarnesen:
Superceded by other PRs

PR #1315, opened 2020-11-19
fix: view as chart has wrong data [DHIS2-9599] [v35]

jenniferarnesen:
This backport doesn't contain any of the refactoring that was done in the implementation on master.  The functionality for getting the adapted ui layout was copied from analytics rather than importing the function. This was to avoid the analytics upgrade (and then having to upgrade dv plugin to keep a single version of analytics).


janhenrikoverland (approved):
LGTM

PR #1318, opened 2020-11-20
test: run cypress ci against live dhis2 instance instead of fixtures

jenniferarnesen:
Switching from fixtures to live instance for now, for 2 reasons:
* Experiencing some issues with generating fixtures with upcoming new tests.
* With live instance, all tests can be run, including mutating tests like creating/deleting dashboards

The dhis2 instance is a dedicated instance on debug.dhis2.org, since dashboard tests are currently heavily reliant on the data, so we don't want people using and changing the instance.

Timeouts were added when initially visiting a dashboard since this seems to take more time with the debug instance than the fixtures.

### DIFF
--- a/cypress/integration/ui/create_dashboard/index.js
+++ b/cypress/integration/ui/create_dashboard/index.js
@@ -28,11 +28,15 @@ const toggleShowMoreButton = () => {
 }
 
 beforeEach(() => {
-    cy.visit('/')
+    cy.visit('/', {
+        timeout: 10000,
+    })
 })
 
 Given('I choose to create new dashboard', () => {
-    cy.get('[data-test="dhis2-dashboard-link-new-dashboard"]').click()
+    cy.get('[data-test="dhis2-dashboard-link-new-dashboard"]', {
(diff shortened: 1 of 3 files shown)


cypress/integration/ui/create_dashboard/index.js
```diff
@@ -28,11 +28,15 @@ const toggleShowMoreButton = () => {
 }
 
 beforeEach(() => {
-    cy.visit('/')
+    cy.visit('/', {
+        timeout: 10000,
```
martinkrulltott, line 32: Consider moving the `timeout: 10000` to a util const for so it can be imported and easily be configured (e.g. to use a different timeout when using stubs) and to prevent repetition.

martinkrulltott (approved):

cypress/integration/ui/create_dashboard/index.js
jenniferarnesen: I will. There will be some more changes coming and I'll address this in an upcoming PR.

PR #1325, opened 2020-11-23
test: add tests for dashboard filter and item context menu

jenniferarnesen:
This PR includes:

New tests:
* added tests for dashboard item context menu
* added tests for dashboard filter

Refactoring:
* ItemFooter is now a functional component (modified this file since I was adding a `data-test` attribute)
* Move common cypress test functions to common.js using cucumber's reusable syntax (using curly braces in step titles)
* Start capturing all data specific to Sierra Leone db backend in a single place.
* Keep cypress selector strings in separate files under `/elements` (e.g. dashboardItem) so that it is easy to update the value if needed


cypress/elements/dashboardItem.js
```diff
@@ -0,0 +1,70 @@
...
+}
+
+const expectChartToExist = itemUid => {
+    if (itemUid) {
+        cy.get(`[data-test="dashboard-item-prog-${itemUid}"]`)
+            .find('.highcharts-container')
+            .should('exist')
+            .and('be.visible')
+    } else {
+        cy.get('.highcharts-container').should('exist').and('be.visible')
```
martinkrulltott: Shouldn't this part somehow be filtered by the `itemUid`? Or how do you know that you've found the right `.highcharts-container` and not just any?

cypress/integration/ui/item_context_menu/item_context_menu.js
```diff
@@ -0,0 +1,41 @@
+import { When, Then } from 'cypress-cucumber-preprocessor/steps'
+import { dashboardItem as itemEl } from '../../../elements/dashboardItem'
```
martinkrulltott: Why rename to `itemEl`?

cypress/elements/dashboardItem.js
```diff
@@ -0,0 +1,70 @@
...
+}
+
+const expectMapToExist = itemUid => {
+    cy.get(`[data-test="dashboard-item-prog-${itemUid}"]`)
+        .find('.dhis2-map-plugin')
+        .should('exist')
+        .and('be.visible')
+}
+
+export const dashboardItem = {
```
martinkrulltott: I like the `action` and `expect` approach, but its a bit confusing since it looks like an object or an instance of a class, but it's not being used as one. Maybe just prefixing the functions with `action` and `expect` will make it easier to understand and use?

martinkrulltott: I mean prefixing and exporting the individual functions, instead of the `export const dashboardItem`

cypress/integration/common/common.js
```diff
@@ -0,0 +1,27 @@
...
+    cy.visit('/', {
+        timeout: 15000,
+    })
+})
+
+Given('I open the {string} dashboard', title => {
+    cy.get('[data-test="dhis2-uicore-chip"]', {
+        timeout: 15000,
+    })
+        .contains(title)
```
martinkrulltott: I would use an explicit version of `contains` to avoid double matches for e.g. `My title` and `My title extended` which would cause the test to fail (since `click` needs a single target)

martinkrulltott: Sorry I didnt realise this was the dashboard title, there can only be one of course.

cypress/integration/common/common.js
```diff
@@ -0,0 +1,27 @@
...
+import { dashboards } from '../../assets/backends'
+
+beforeEach(() => {
+    cy.visit('/', {
+        timeout: 15000,
+    })
+})
+
+Given('I open the {string} dashboard', title => {
+    cy.get('[data-test="dhis2-uicore-chip"]', {
```
martinkrulltott: Can this be targeted more specifically by adding a custom data-test prop? `dhis2-uicore-chip` could be used in other places?

martinkrulltott: `view-dashboard-title`?

cypress/integration/ui/dashboard_filter/dashboard_filter.js
```diff
@@ -0,0 +1,26 @@
...
+
+const OPTIONS = { timeout: 15000 }
+
+When('I add a {string} filter', () => {
+    cy.contains('Add filter').click()
+
+    cy.get('[data-test="undefined-button-pe"]').click()
+
+    cy.get('[data-test="dhis2-uicore-transfer-sourceoptions"]')
+        .contains('Last 6 months')
```
martinkrulltott: For scalability I would move this out to a const e.g. `TEST_PERIOD` to be used both for the action and assertion below.

cypress/integration/ui/item_context_menu/item_context_menu.js
martinkrulltott: Name-wise to me I get the impression that this is the same element that keeps being reused (i.e. as if it was an instance of a class), but it seems like this represents all dashboard items and the item itself is targeted by providing an id. I think these two concepts are clashing a bit from a readability perspective.
I suggest either naming it something more generic, to not get the impression of being a specific unique element.
Or having it receive the itemUid as a prop, like `itemEl(chartItemUid).action.clickMenuButton()`

martinkrulltott: In the second alternative I'd probably instantiate `dashboardItem` twice instead of passing the id all the time, like `const chartItemEl = dashboardItem (dashboards.Delivery.items.chart.uid)` and then use it like `chartItemEl.action...`

cypress/integration/ui/dashboard_filter/dashboard_filter.js
```diff
@@ -0,0 +1,26 @@
+import { When, Then } from 'cypress-cucumber-preprocessor/steps'
+
+const OPTIONS = { timeout: 15000 }
+
+When('I add a {string} filter', () => {
+    cy.contains('Add filter').click()
+
+    cy.get('[data-test="undefined-button-pe"]').click()
```
martinkrulltott: `undefined-button-pe`, why the `undefined`?

cypress/integration/ui/dashboard_filter/dashboard_filter.js
```diff
@@ -0,0 +1,26 @@
+import { When, Then } from 'cypress-cucumber-preprocessor/steps'
+
+const OPTIONS = { timeout: 15000 }
```
martinkrulltott: I think this can be moved to a global util prop to be reused, e.g. `.utils/options.js` with `export const extendedTimeout = ....`

cypress/elements/dashboardItem.js
jenniferarnesen: Removed the else - not in use.

cypress/integration/ui/dashboard_filter/dashboard_filter.js
jenniferarnesen: 👍

cypress/elements/dashboardItem.js
jenniferarnesen: Things have changed, so no longer relevant

cypress/integration/common/common.js
jenniferarnesen: Done. Using 'dashboard-chip'

cypress/integration/ui/dashboard_filter/dashboard_filter.js
jenniferarnesen: That's coming from Analytics. But not using this selector anymore.

cypress/elements/dashboardFilter.js
```diff
@@ -0,0 +1,6 @@
+export const filterBadgeSel = '[data-test="dashboard-filter-badge"]'
```
martinkrulltott: `..Sel` not `..El` ? 🤔

martinkrulltott: El = element,
Sel = selector?
But the file is in /elements/....

cypress/elements/dashboardItem.js
```diff
@@ -0,0 +1,15 @@
+export const chartSel = '.highcharts-container'
+export const chartSubtitleSel = '.highcharts-subtitle'
+export const chartXAxisLabelSel = '.highcharts-xaxis-labels'
+export const mapSel = '.dhis2-map-plugin'
+export const tableSel = '.pivot-table-container'
+
+export const itemDetailsSel = '[data-test="dashboarditem-footer"]'
+
+export const getDashboardItem = itemUid =>
```
martinkrulltott: would probably be good to differentiate these chainable functions (that returns a `cy` object) in a way that makes it clear that they are indeed chainable. I have the same issue in DV, but haven't come up with a good way to do it yet.

However in this case, where all the function does is `cy.get(id)` I think it would be better to not use a function, as it doesn't really add much functionality. If this is ok here then really it should be done for all `Sel`'s above too 🤷‍♂️  Or is this because of the concat selector `dashboarditem-${itemUid}`? If the `itemUid` is truly unique you could just a wildcard selector instead, like I've got `getBySelLike` in DV.
Or if you insist on prefixing the id with `dashboarditem-` (which shouldn't really be necessary) you could create a new command that does this for you.

martinkrulltott (approved):

cypress/elements/dashboardFilter.js
jenniferarnesen: Changed dir name to selectors

cypress/elements/dashboardItem.js
jenniferarnesen: I'll revisit this in an upcoming PR.

PR #1333, opened 2020-11-26
chore: replace travis with github action

jenniferarnesen:
Based on the template workflow "verify-app". The cypress action has been moved to the verify action.


varl:
Looks good!

Why not merge the Cypress action into the unified one? It allows you to save an `install` command and re-use the cached `node_modules` in addition to control when the e2e tests are run. For example, there is no need to run expensive e2e tests if the cheap unit tests fail for example.

jenniferarnesen:
> Looks good!
> 
> Why not merge the Cypress action into the unified one? It allows you to save an `install` command and re-use the cached `node_modules` in addition to control when the e2e tests are run. For example, there is no need to run expensive e2e tests if the cheap unit tests fail for example.

Makes sense!

amcgee:
This seems to have created a new repository @d2-ci/dashboard-app, I think the name in package.json needs to be updated

varl (approved):

PR #1334, opened 2020-11-26
feat: hide app title where applicable (DHIS2-9605)

tomzemp:
Hide title in AppItem/Item if specified in app's manifest


src/components/Item/AppItem/Item.js
```diff
@@ -39,19 +39,34 @@ const AppItem = ({ dashboardMode, item, itemFilters }, context) => {
         )
     }
 
+    const hideTitle =
```
jenniferarnesen, line 42: It should be possible to write it like this:

`appDetails?.settings?.dashboardWidget?.hideTitle && dashboardMode !== EDIT`
(need to import EDIT from dashboardModes)

tomzemp: Sounds good. Wasn't sure how you felt about ES2020 features 😄 

Updated: 

https://github.com/dhis2/dashboard-app/pull/1334/commits/a1cabd70c3e9eeb050ab2b2e03cd1efafeb3bab9

benguaraldi: @jenniferarnesen If we use 
```
appDetails?.settings?.dashboardWidget?.hideTitle && dashboardMode !== EDIT
```
it looks like that will cut out some browser support:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining

Is DHIS2 supposed to support any of those browsers?  e.g., mobile Firefox?

jenniferarnesen: @benguaraldi 
We use babel, which transforms the code so that it is compatible with older browsers. So the optional chaining syntax is fine to use here.

That said, as of 2.33, DHIS2 no longer officially supports IE11. Just Chrome, FF, Safari and Edge.  This could probably be better communicated!

btw, what version are you planning these changes to be backported to? Isn't Datim on a slightly older version?

src/components/Item/AppItem/Item.js
```diff
@@ -39,19 +39,31 @@ const AppItem = ({ dashboardMode, item, itemFilters }, context) => {
...
+        appDetails?.settings?.dashboardWidget?.hideTitle &&
+        dashboardMode != EDIT
```
jenniferarnesen: Oops, missed this: use strict equality checks: `!==`

jenniferarnesen (approved):

src/components/Item/AppItem/Item.js
benguaraldi: @jenniferarnesen Ah, got it!

Right now, we're focused on putting new features in trunk so they will eventually be available to us.  Most of these things are nice-to-haves for us, so I wasn't planning to try to convince folks to let us backport them, as I know that's generally not done with features.

However, if backporting were allowed, we'd might go back to 2.33 (as that's what we're running) or even 2.32, as that will make our Dashboard Information Widget work better on those older versions.

jenniferarnesen: ok, we'll focus on trunk then. Backports can be really messy sometimes so yes we're trying to avoid them. And 2.32 has gone out of official support.

src/components/Item/AppItem/Item.js
tomzemp: fixed

jenniferarnesen (approved):

jenniferarnesen:
Merged in a separate PR in order to run tests

PR #1339, opened 2020-12-01
chore: refactor Visualization item to separate Item from Plugin

jenniferarnesen:
This is a refactoring PR. There should be no visual or functional changes.  This work is the foundation for the fullscreen implementation (which will add new functionality to the Item component)

What was done:
* *Visualization/Item* was getting very big and had too many responsibilities. It has been separated into components:
* * *Item*: deals with general things related to item - the menu, the expansion of the footer, the header.
* * *Visualization*: deals with displaying the visualization - gets the config for the active type, applies the filter, and renders the correct plugin.
* A new cypress test has been added to confirm that charts are not unnecessarily being reloaded when an item is moved in edit mode (this has been a previous regression, and may happen again with future changes to this item)

Other notes:
* You might be wondering why I didn't switch to functional components with React hooks. Believe me, I tried. But I was continuously getting the reloading problem mentioned above, so I put that aside for now.
* The edit_dashboard (renamed from create_dashboard) cypress test file is 90% the same as before, just with the addition of the new test at the bottom of the file.


src/components/Item/VisualizationItem/Visualization/__tests__/Visualization.spec.js
```diff
@@ -0,0 +1,134 @@
...
+test('renders a VisPlugin when activeType is CHART', () => {
+    const store = {
+        visualizations: { rainbowVis: { rows: [], columns: [], filters: [] } },
+    }
+    const { container } = render(
+        <Provider store={mockStore(store)}>
+            <Visualization
+                item={{
+                    id: 'rainbow',
+                    type: 'REPORT_TABLE',
```
martinkrulltott: Why doesn't this and the next test have the same `type` as `activeType`? 
`type: 'REPORT_TABLE'` and `activeType="CHART"` 🤔

src/components/Item/VisualizationItem/Visualization/__tests__/Visualization.spec.js
```diff
@@ -0,0 +1,134 @@
...
+
+jest.mock('@dhis2/data-visualizer-plugin', () => 'VisualizationPlugin')
+jest.mock('../MapPlugin', () => 'MapPlugin')
+jest.mock('../DefaultPlugin', () => 'DefaultPlugin')
+
+const mockStore = configureMockStore()
+
+test('renders a MapPlugin when activeType is MAP', () => {
+    const store = {
+        visualizations: { rainbowVis: { rows: [], columns: [], filters: [] } },
```
martinkrulltott: `visualizations: { rainbowVis: { rows: [], columns: [], filters: [] } },` is repeated 5 times. Set it as e.g. `const DEFAULT_STORE_WITH_1_ITEM = ...` and use `const store = DEFAULT_STORE_WITH_1_ITEM` inside the tests?

src/components/Item/VisualizationItem/Visualization/getFilteredVisualization.js
```diff
@@ -0,0 +1,42 @@
...
+    }
+
+    // deep clone objects in filters to avoid changing the visualization in the Redux store
+    const visRows = visualization.rows.map(obj => ({ ...obj }))
+    const visColumns = visualization.columns.map(obj => ({ ...obj }))
+    const visFilters = visualization.filters.map(obj => ({ ...obj }))
+
+    Object.keys(filters).forEach(dimensionId => {
+        if (filters[dimensionId]) {
+            let dimensionFound = false
```
martinkrulltott, line 13: There's really a lot going on here... I'm generally not a fan of `xxFound = true` and the three nested `forEach` loops makes it quite hard to read. I'm sure this could be solved in a simpler way by e.g. checking if the `length` of the matched items is 0 and then push to `visFilters`. But I'd need proper test data to write it and TBH I'm not sure if the readability would actually be increased. 
However if you've got time I'd consider refactoring this and if not at least change the prop name `obj` to something that gives semantic meaning and include a comment with the object structure of the `visualization` props so one knows what to expect.

martinkrulltott (approved):

src/components/Item/VisualizationItem/Visualization/getFilteredVisualization.js
jenniferarnesen: Agreed that this function has a lot going on. I'll add this to the tech debt list.

src/components/Item/VisualizationItem/Visualization/__tests__/Visualization.spec.js
jenniferarnesen: This is simulating the "View as" feature of dashboards. E.g., viewing a chart as a table, or a table as a chart. I've changed the test name so that it's hopefully more clear.

src/components/Item/VisualizationItem/Visualization/__tests__/Visualization.spec.js
jenniferarnesen: Done.

PR #1358, opened 2020-12-07
feat: add fullscreen option to visualization items [DHIS2-9879]

jenniferarnesen:
image

image

Todo:
* Remove fullscreen icon from map (separate implementation)
* Get fullscreen icon from ui-icons instead of having the svg code in the app

Upcoming refactor:
* the current solution relies on non-react use of `document.querySelector`. For best practice, this should instead be done with refs, but using refs would have required a somewhat big refactor. I would prefer to do this refactor separately since the testing scope is larger.

Other:
* No cypress tests added since it isn't currently possible to test the fullscreen api in cypress. I'll be keeping an eye on if/when this gets implemented in cypress (https://github.com/cypress-io/cypress/issues/7776)
* Minor refactor: move check for whether to show ItemHeaderButtons (pluginIsAvailable) out of ItemHeaderButtons and to Item, so that the logic doesn't get too complex in ItemHeaderButtons (would have been first checking pluginIsAvailable, then checking the fullsreen prop)


src/components/Item/VisualizationItem/Item.js
```diff
@@ -60,6 +63,19 @@ export class Item extends Component {
         this.setState({
             configLoaded: true,
         })
+
+        const el = document.querySelector(this.itemDomElSelector)
+        if (el?.requestFullscreen) {
+            el.onfullscreenchange = this.handleFullscreenChange
+        }
```
edoardo, line 86: I'm wondering if we should handle the error case with `onfullscreenerror`

edoardo (approved):

PR #1360, opened 2020-12-09
chore: upgrade @dhis2/cli-utils-cypress and @dhis2/cypress-commands to 4.0.0

jenniferarnesen:
Triggers upgrade to cypress@6


cypress/support/index.js
```diff
@@ -1,7 +1,7 @@
-import { registerCommands } from '@dhis2/cypress-commands'
+import { enableAutoLogin } from '@dhis2/cypress-commands'
 
 import { enableNetworkShim, loginAndPersistSession } from './server'
 
-registerCommands()
+enableAutoLogin()
 enableNetworkShim()
 loginAndPersistSession()
```
martinkrulltott: For DV I removed `loginAndPersistSession` completely, but maybe its needed here for testing stubs?

martinkrulltott (approved):

cypress/support/index.js
jenniferarnesen: You're right. Removed that file and some more code that is handled by enableAutoLogin.

PR #1400, opened 2020-12-17
feat: responsive control bar and title bar [DHIS2-10138]

jenniferarnesen:
Partially implements DHIS2-10138
This PR addresses responsive changes in the view control bar and view title bar. Includes
* spacing around chips, dashboard search field, against edge, chip scroll bars
* remove action buttons (Edit, Share, Add filter)
* remove "+" button (new button)

*Width <=480 and  collapsed control bar

* Width <=480 and expanded control bar

Implementation notes:
* decoupled the View control bar from Edit control bar, since Edit mode doesn't need all the dnd functionality
* For js part of the implementation, a new react hook was added that listens to window resizes. The components that need to make responsive changes can then use that hook.
* The ChevronUp and ChevronDown icons are currently copied into this repo, but when we upgrade to ui6, then we'll switch to using the icons from there.
* ControlBar was simplified since it is no longer used for the Edit bar (no `editMode` needed)
* Where it made sense, I switched from inline css to css classes
* The "Show more/less" button on the control bar is now an icon instead of text. Accessibility changes were made to adapt to this, like using a <button> element and adding aria-label. Also added a tooltip, suggested by Joe.


cooper-joe:
Looks good generally, though there are a few things that do not follow the spec (for example, [the height of the expanded control bar should be 100vh - 32px](https://docs.google.com/document/d/1zhqRu_2wE8JpFJBbmS5b2e5Jo5rcSo-OANnek5qLcpI/edit#heading=h.c0sz4g2rmzzk), and there a couple of bugs with scrolling inside the control bar - but I'm not sure if these are out of scope of this PR. The PR description states that this partially implements the spec, so I'm unsure which parts should be reviewed/evaluated against the spec.

Should I evaluate against the spec now, or wait until a final PR that would implement DHIS2-10138 fully?

jenniferarnesen:
> Looks good generally, though there are a few things that do not follow the spec (for example, [the height of the expanded control bar should be 100vh - 32px](https://docs.google.com/document/d/1zhqRu_2wE8JpFJBbmS5b2e5Jo5rcSo-OANnek5qLcpI/edit#heading=h.c0sz4g2rmzzk), and there a couple of bugs with scrolling inside the control bar - but I'm not sure if these are out of scope of this PR. The PR description states that this partially implements the spec, so I'm unsure which parts should be reviewed/evaluated against the spec.
> 
> Should I evaluate against the spec now, or wait until a final PR that would implement DHIS2-10138 fully?

@cooper-joe  I was aware of the controlbar height not meeting the spec - that will take a little work.  I wasn't aware of the scrolling issues.  Could you add some comments to https://jira.dhis2.org/browse/DHIS2-10138, and I'll address these separate PRs?

cooper-joe:
>  I wasn't aware of the scrolling issues. Could you add some comments to https://jira.dhis2.org/browse/DHIS2-10138, and I'll address these separate PRs?

I was just trying to capture the bug to report, but I realize it's actually  to do with the control bar not being the correct height (as noted above), so I was getting confused about which area was scrolling. So I think it's not a bug, it's just not completed yet. 👍

edoardo (approved):

PR #1431, opened 2021-01-06
feat: responsive item grid in view and edit mode [DHIS2-10138]

jenniferarnesen:
Implements the responsive dashboard grid in View and Edit mode.
* The responsive view has a single breakpoint at 480px, at which point the dashboard will change to a single column format (view mode only - edit mode will not be available at <480px)
* We attempt to preserve the original aspect ratio of the grid items by making the assumption that the dashboard creator used a desktop with an approximately 1200px wide viewport. Thus, the height will be scaled down or up to match keep the ratio.
* Visualization items have a minimum height of about 320px
* Non visualizations will keep their original pixel height.
* padding between dashboard items increases from 10px to 16px on small screen

Refactoring:
* move `gridUtil` to modules folder (and move some functions into separate modules (smallScreen, getGridItemDomId) to reduce scope of gridUtil)
** these functions in gridUtil haven't changed: `getPrintTitlePageItemShape `, `getPageBreakItemShape `, `hasShape`, `withShape`, `getShape`, `isNonNegativeInteger`
* create separate `ViewItemGrid` and `EditItemGrid` components (used to be ItemGrid with editMode flag) since functionality and grid props very different
* start replacing `windowHeight` redux prop with the `useWindowDimensions` hook
* reduce duplication in `PrintItemGrid` and `PrintLayoutItemGrid` by breaking out a `StaticGrid` component with the common functionality


src/components/Item/VisualizationItem/Item.js
```diff
@@ -167,6 +195,13 @@ export class Item extends Component {
         )
     }
 
+    getAvailableWidth = () => {
+        const rect = document
+            .querySelector(this.itemDomElSelector)
+            ?.getBoundingClientRect()
+        return rect && rect.width - ITEM_CONTENT_PADDING * 2
```
edoardo, line 202: Could this end up returning a negative value?
What would happen in that case?

edoardo (approved):

src/components/Item/VisualizationItem/Item.js
jenniferarnesen: From my testing, the minimum possible width that the items can be set to is 70px. So the smallest number this fn would ever return is 62.

PR #1441, opened 2021-01-12
refactor: replace old context with hooks

jenniferarnesen:
The main goal of this PR is to remove old context, which had properties: d2, baseUrl, i18n, dataEngine
* `i18n` was not in use anywhere
* `baseUrl` is replaced by a hook from app-runtime
* `d2` is replaced by a hook from app-runtime-adapter-d2
* `dataEngine` is replaced by a hook from app-runtime

To use the hooks, some components were converted to functional components. 

The reason I am doing this refactoring now is because I'll be using a hook in the responsive dashboard functionality.

### Notes about changes to the components

**EditBar:
AppItem:
ListItem:
FilterDialog:
ItemSelector:
ViewTitleBar:
ItemFooter:**
- use hook instead of old context

**App:**
- remove old context
- switch to functional component
- component markup has not changed (apparent changes are only indentation)

**MessagesItem:**
- switch to functional component
- use hooks instead of old context (old context baseUrl was undefined so this was actually a bug)
- component markup has not changed (apparent changes are only indentation)

**Visualization Item:**
- removed old context d2, which was not in use

**DefaultPlugin (lots of changes related to switching to functional component with hooks):**
- itemFilters property was not impacting the component in any way and was removed
- technique with useRef in order to compare prev props with current props when determining whether to reload the plugin (this is to avoid unnecessary flashing from reloading the visualization)
- `pluginIsAvailable` is moved to the parent component so it doesn't have to be checked numerous times in this component
- TODO: add cypress test coverage for this component (Map, EV, ER items)

**MapPlugin:**
- check `pluginIsAvailable` before rendering DefaultPlugin

**Visualization:**
- keep as class due to complex memoization code, so use `D2Shim` to remove old context
- check `pluginIsAvailable` before rendering DefaultPlugin

**CategorizedMenuGroup:**
- switch to functional component
- use baseUrl hook instead of d2 since the old d2 was just used to get the baseUrl
- component markup has not changed (apparent changes are only indentation)


src/components/ItemSelector/ItemSelector.js
```diff
@@ -25,7 +24,7 @@ const ItemSelector = ({}, context) => {
 
         const filterStr = filter ? `/${filter}` : ''
 
-        context.d2.Api.getApi()
+        d2.Api.getApi()
```
jenniferarnesen, line 27: @edoardo Could dataEngine be used here instead?

src/components/ControlBar/EditBar.js
```diff
@@ -27,57 +29,62 @@ import { apiFetchDashboard } from '../../api/dashboards'
...
-        dashboard: undefined,
-        confirmDeleteDialogOpen: false,
-        redirectUrl: undefined,
-    }
+const EditBar = props => {
+    const { d2 } = useD2({})
+    const dataEngine = useDataEngine()
+    const [translationDlgIsOpen, setTranslationDlgIsOpen] = useState(false)
+    const [dashboard, setDashboard] = useState(undefined)
+    const [confirmDeleteDlgOpen, setConfirmDeleteDlgOpen] = useState(false)
```
edoardo: Minor: keep the same format as `translationDlgIsOpen`?

src/components/Item/VisualizationItem/Visualization/DefaultPlugin.js
```diff
@@ -1,111 +1,64 @@
...
-        this.pluginCredentials = pluginCredentials(this.d2)
-
-        if (this.pluginIsAvailable()) {
-            pluginManager.load(this.props.item, this.props.visualization, {
-                credentials: this.pluginCredentials,
-                activeType: this.props.activeType,
-                options: this.props.options,
+    useEffect(() => {
+        if (shouldPluginReload()) {
+            unmount(item, prevActiveType.current)
```
edoardo, line 34: If I understand correctly `prevItem.current` will always be the same as `item`, so it doesn't matter which one is used here.
Correct?

src/components/Item/VisualizationItem/Visualization/Visualization.js
```diff
@@ -79,16 +81,20 @@ class Visualization extends React.Component {
                                 <LoadingMask />
                             </div>
                         )}
-                        <VisualizationPlugin
-                            d2={this.context.d2}
-                            visualization={this.memoizedGetFilteredVisualization(
-                                pluginProps.visualization,
-                                pluginProps.itemFilters
+                        <D2Shim d2Config={{}}>
```
edoardo, line 84: Does this return the same `d2` instance if the config passed is different?
Ie. here there is no `schemas` passed, so is this causing the all schemas request to fire?

src/AppWrapper.js
```diff
@@ -49,7 +49,7 @@ const AppWrapper = () => {
                         }
                         return (
                             <WindowDimensionsProvider>
-                                <App d2={d2} dataEngine={dataEngine} />
+                                <App />
```
edoardo, line 52: Should we pass empty `schemas` to the `D2Shim` above to avoid the schema request?

src/components/ItemSelector/ItemSelector.js
edoardo: I think it should be possible, although with a hacky way of dealing with that URL.

edoardo (approved):

src/components/Item/VisualizationItem/Visualization/Visualization.js
jenniferarnesen: If i understand the code correctly, `thed2` is a singleton and if it has previously been instantiated, then the initialization code does not run. https://github.com/dhis2/app-runtime-adapter-d2/blob/master/src/useD2.js#L43
So this should be safe. @amcgee , confirm?

And actually, this branch requests the schemas only once, while in prod, they are requested twice.

src/components/ControlBar/EditBar.js
jenniferarnesen: Done

src/AppWrapper.js
jenniferarnesen: As mentioned above, I don't think this is needed.

src/components/Item/VisualizationItem/Visualization/Visualization.js
edoardo: Do we need any schema at all?

src/AppWrapper.js
edoardo: But I don't think `d2` does anything by default with schemas.
Shouldn't we pass `[]` (or any schema we actually need) to avoid all the schemas from being fetched?

src/components/Item/VisualizationItem/Visualization/DefaultPlugin.js
jenniferarnesen: Right. So the goal here is that if the user changes the active type (i.e., EV -> ER), then the EV plugin should unmount, and the ER plugin should load.

src/components/Item/VisualizationItem/Visualization/Visualization.js
jenniferarnesen: Yes, we need the schemas. We tried removing them in 2.35, and the result was that all schemas were fetched. Then we added back only some of them, but things like d2-ui components were broken without them all. (e.g. organisationUnits)

jenniferarnesen: Try this in the dashboard repo:  `git show 4e3c83ec`

src/AppWrapper.js
jenniferarnesen: The D2Shim receives the list of schemas on line 44 of AppWrapper.js

Then, the d2 singleton (accessed via `useD2` and other `D2Shim`s) has the schemas loaded in it.

edoardo: Right, sorry my bad, the relevant code was hidden in the diff here so I missed the schema part.

src/components/Item/VisualizationItem/Visualization/DefaultPlugin.js
edoardo: 👍

src/components/Item/VisualizationItem/Visualization/Visualization.js
edoardo: 👌

PR #1456, opened 2021-01-19
test: dashboard api tests

jenniferarnesen:
Covers:
* fetch all dashboards
* create dashboard
* fetch single dashboard
* delete dashboard

Right now this PR is more of a proof-of-concept than final product (I can make the test code more efficient and succinct).

The idea for having these tests is two-fold:
* isolate when a problem in the dashboard app is caused by the api
* provide the api team with suites of tests that can be run during development to confirm that their changes do not break core apps ("contract tests")

I haven't addressed how to choose a particular api version to run, and also how the api team would run these tests in their environment. But I like the idea of the tests coexisting with the app.

Also for consideration:
* when a request comes from a lib (e.g., analytics), where should the api test be located? I can see arguments for both putting them in the app, or putting them with the lib. There would be a lot more duplication if it is in the app, but also is perhaps easier to relate to for the api team.


amcgee:
@vilkg what do you think about pulling app tests into the core build/test pipeline?

vilkg:
Generally, I think that API tests should be co-located with the backend. We could have a pipeline that pulls the tests from all the apps, but IMO it should be separate from the core build/test pipeline we run on PRs. We could look into consumer-driven contract tests as well, which would require a bigger discussion and common standards for defining those contracts across apps.

jenniferarnesen:
> Generally, I think that API tests should be co-located with the backend. We could have a pipeline that pulls the tests from all the apps, but IMO it should be separate from the core build/test pipeline we run on PRs. We could look into consumer-driven contract tests as well, which would require a bigger discussion and common standards for defining those contracts across apps.

The reason to have the tests co-located with the app is that it is the app developer who knows what apis the app needs. And in that case it makes sense to lower the bar for writing such tests, by using the same repos and technologies as the app itself. 

If these shouldn't be run on the core build/test pipeline for backend PRs, where do you see them being run?

vilkg:
> The reason to have the tests co-located with the app is that it is the app developer who knows what apis the app needs. And in that case, it makes sense to lower the bar for writing such tests, by using the same repos and technologies as the app itself.

yeah, I understand that this approach is useful and we should certainly have a bigger discussion about it. I suggest to have it at the test automation meeting. Running these tests on backend PRs could result in numerous failures that the backend wouldn't have any control over, like: 
- expectations you make on the metadata and data being present (I understand you depend on the SL db in your tests)
- tests being flaky/outdated/wrong, that would require involvement from the app devs to resolve failures

so until we figure out a reliable solution and common format, I would suggest having a separate pipeline on, say, Jenkins.

PR #1459, opened 2021-01-21
feat: set expanded control bar height for small screens

jenniferarnesen:
Fixes in this PR:
* Small screen expanded controlbar: the search input field should stretch to the whole width
* Small screen tapping “search icon” in the control bar, search input should focus
* Small screen expanded controlbar:  adjust "search icon"
* Small screen expanded controlbar should take all the height - 32px
* Vertical scrolling in the expanded controlbar should work
 

**Control bar height (100vh - 32px)**
image

**Search input focus and whole width**
image

**Vertical scroll**
image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,10 @@
+# [31.8.0](https://github.com/dhis2/dashboard-app/compare/v31.7.1...v31.8.0) (2021-01-28)
+
+
+### Features
+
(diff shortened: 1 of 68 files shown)


Sharmyn28 (approved):

cooper-joe (approved):
LGTM 👍 

I see we get the double-scrollbars when the control bar is expanded (scrollbar for controlbar, scrollbar for dashboard content). I think this might be inevitable, but I'll do some extra testing when this is on play.

PR #1461, opened 2021-01-21
fix: trigger event to hide tooltip when click wasn't on button

jenniferarnesen:
Problem: the tooltip for the Show more button (indicated by Chevron up/down) didn't get removed when clicking on the button. This is because when the control bar expands or collapses, the `mouseout` event didn't get triggered for the tooltip.


janhenrikoverland (approved):
LG

PR #1473, opened 2021-01-27
feat: responsive - redirect to view when url of /edit or /new manually entered [DHIS2-10138]

jenniferarnesen:
If the app is displayed in small screen, then edit and new are not supported. If the user manually enters the /edit or /new route, then the app will redirect back to the view mode of the same dashboard that was being viewed.

Note that if the user was already in /edit or /new, and then changed the screen to <=480px, then they won't get redirected, but a message will be displayed about not being able to edit in small screen.

### DIFF
--- a/cypress/integration/ui/responsive_dashboard/responsive_dashboard.js
+++ b/cypress/integration/ui/responsive_dashboard/responsive_dashboard.js
@@ -1,4 +1,7 @@
 import { When, Then } from 'cypress-cucumber-preprocessor/steps'
+import { EXTENDED_TIMEOUT } from '../../../support/utils'
+import { chartSel } from '../../../selectors/dashboardItem'
+import { dashboardTitleSel } from '../../../selectors/titleBar'
 
 const TEST_DASHBOARD_TITLE = 'TEST_DASHBOARD_TITLE'
 
@@ -68,5 +71,34 @@ Then('my changes are still there', () => {
         const val = $input.val()
 
         expect(val).to.match(re)
+    })
+})
+
+// Scenario: I change the url to new while in small screen
+When('I change url to new', () => {
+    const url = `${Cypress.config().baseUrl}/#/new`
+    cy.window().then(win => {
+        win.location.assign(url)
+        cy.wait(2000) // eslint-disable-line cypress/no-unnecessary-waiting
+    })
+})
+
+Then('the {string} dashboard displays in default view mode', title => {
+    cy.location().should(loc => {
+        expect(loc.hash).to.equal('#/')
+    })
+
+    cy.get(dashboardTitleSel).should('be.visible').and('contain', title)
+    cy.get(chartSel, EXTENDED_TIMEOUT).should('exist')
+})
+
+// Scenario: I change the url to 'edit' while in small screen
+When('I change url to edit', () => {
+    cy.location().then(loc => {
+        const url = `${loc.href}/edit`
+        cy.window().then(win => {
+            win.location.assign(url)
+            cy.wait(2000) // eslint-disable-line cypress/no-unnecessary-waiting
+        })
     })
 })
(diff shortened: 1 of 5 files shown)


cooper-joe (approved):

cypress/integration/ui/responsive_dashboard/responsive_dashboard.js
```diff
@@ -70,3 +73,32 @@ Then('my changes are still there', () => {
         expect(val).to.match(re)
     })
 })
+
+// Scenario: I change the url to new while in small screen
+When('I change url to new', () => {
+    const url = `${Cypress.config().baseUrl}/#/new`
+    cy.window().then(win => {
+        win.location.assign(url)
+        cy.wait(2000) // eslint-disable-line cypress/no-unnecessary-waiting
```
martinkrulltott, line 82: Why wait 2s? Can you wait for an element to render instead? (e.g. `expect`)

src/components/Dashboard/Dashboard.js
```diff
@@ -59,10 +64,22 @@ export const Dashboard = ({
...
     setWindowHeight,
 }) => {
+    const { width } = useWindowDimensions()
+    const [redirectUrl, setRedirectUrl] = useState(null)
+
     useEffect(() => {
         setHeaderbarVisibility(mode)
     }, [mode])
 
+    useEffect(() => {
```
martinkrulltott, line 74: Is it necessary to put the redirect logic in a `useEffect(..., [mode])`? If you're only catching the case of directly typing the url then the mode isn't changing?

src/components/Dashboard/Dashboard.js
```diff
@@ -90,6 +107,10 @@ export const Dashboard = ({
         )
     }
 
+    if (redirectUrl) {
+        return <Redirect to={redirectUrl} />
+    }
+
     if (mode === NEW) {
```
martinkrulltott, line 114: Not that it matters for the code execution, but I would chain these `if`s to use `else if` to make it clear that only one of them will be triggered.

martinkrulltott: and wrap the final `return dashboardMap[mode]` in an `else`

martinkrulltott: actually, I'm not a fan of multiple return statements, so I'd probably extract the return value to a prop (e.g. `result`) and assign it in the different cases and just use a single `return result` in the end, but that might be too much out of scope for this PR 🤷‍♂️

martinkrulltott: ... changed my mind, multiple returns can be ok I guess. But the `else if` would be nice either way 🙂

cypress/integration/ui/responsive_dashboard/responsive_dashboard.js
martinkrulltott: Ah, no, because the redirect doesn't render anything. But what happens when you don't wait?
Or could you expect the url to be a certain value?

martinkrulltott (approved):
Looks good to me. I've never used `redirect`, so I'm a bit curious how it works in Cypress, hence my comment. But I guess you added the `wait` for a reason 🙂 
Anyway, I'd probably add a negative test case to make sure the redirect doesn't trigger when a large screen is used. Unless you think this is already covered by the existing test suite.

src/components/Dashboard/Dashboard.js
jenniferarnesen: Just tried a couple of other ways, but none of them worked. The mode actually does change when the user types in "edit" or "new".

src/components/Dashboard/Dashboard.js
jenniferarnesen: I think this is one of those rare cases where multiple returns is ok. There have been several bad bugs (crashes) from this code before it got to its current reincarnation.  So I kind of want to leave as is for now.

PR #1484, opened 2021-01-29
fix: fullscreen param to map plugin resize

jenniferarnesen:
map plugin resize function now accepts an `isFullscreen` param


turban (approved):
This looks correct. Please notify me when merged to master, and I'll do some extra testing on play dev.

PR #1501, opened 2021-02-02
fix: change order of save process to avoid memory leaks

jenniferarnesen:
Implementations in this PR:
* change order of save process to avoid memory leaks
* add Alert in case saving fails, and remain on the Edit page (related to this, no longer wrap the api requests with try/catch, and let the component handle the error)
* New cypress tests that check several edge case/error scenarios (I'm loving `cy.intercept`)
* Other minor cleanup in the cypress tests

## More info about the new save process:

Sometimes saving a dashboard causes console warnings about setting state on an unmounted component, causing memory leaks.  (test this locally, by freshly loading the app, click Edit, make a change, then click Save)

Previously, the order of events was:
1. save dashboard (in edit mode)
2. redux actions to clear edit mode
3. redirect to the dashboard in view mode
4. load the new/changed dashboard and more redux actions to set the state 

What is happening is that redux state was being changed prior to Redirect, causing the edit view to unmount and remount. Then once the Redirect has taken place, the still-mounted edit components were trying to set state.

Now, the process is like this:
1. save dashboard
2. fetch the updated list of dashboards
3. redirect to view
4. redux clear the edit mode
5. load the new/changed dashboard and more redux actions to set the state

## Alert on failed save
In addition, if the save request fails, then it is caught and an Alert is shown. The user remains in Edit view (should discuss this new feature with Joe/Jan - maybe user should still be redirected to View if the error was due to lack of access?)

You can test the bad behaviour by trying to save the "Cases Malaria" dashboard.  (this is probably caused by some weird data set up of the access)

image


src/actions/dashboards.js
```diff
@@ -100,15 +96,12 @@ export const tStarDashboard = (id, isStarred) => async (
...
-    const onError = error => {
-        console.log('Error (apiStarDashboard): ', error)
-        return error
-    }
     try {
         await apiStarDashboard(dataEngine, id, isStarred)
         return onSuccess(id)
     } catch (err) {
-        return onError(err)
+        console.log('Error (apiStarDashboard): ', err)
```
martinkrulltott: Use `console.error` instead of `log`?

src/components/ControlBar/EditBar.js
```diff
@@ -50,9 +56,16 @@ const EditBar = props => {
...
-            setRedirectUrl(`/${newId}`)
-        })
+        try {
+            props
+                .onSave()
+                .then(newId => {
+                    setRedirectUrl(`/${newId}`)
+                })
+                .catch(() => failureAlert.show())
+        } catch (e) {
```
martinkrulltott: why receive `(e)` if it's not used?
I'm surprised this wasn't caught by your linter...

martinkrulltott (approved):
Tests LGTM 👍 
Nice use of `cy.intercept`

cooper-joe:
I think this is a great suggestion, it's much clearer to the user what's going on when there is an issue. Most importantly, a user is never unsure about whether they have saved.

> In addition, if the save request fails, then it is caught and an Alert is shown. The user remains in Edit view (should discuss this new feature with Joe/Jan - maybe user should still be redirected to View if the error was due to lack of access?)

I do not think the app should ever redirect after a failed save, for whatever reason. Doing so would programatically discard changes. This would be unexpected data loss. To avoid this, it should always be the user's responsibility to click "Exit without saving". _With user intent_ should be the only situation we discard unsaved changes.

i18n/en.pot
```diff
@@ -29,6 +29,9 @@ msgid ""
 "this dashboard?"
 msgstr ""
 
+msgid "Failed to save the dashboard"
```
cooper-joe, line 32: It would be useful to be more specific, or suggestive, of a possible fix. I'm thinking `Failed to save dashboard. You might be offline or not have access to edit this dashboard.` Are there other reasons this error might end up being displayed? Perhaps "does not exist anymore"?

jenniferarnesen: Does not exist could be possible, though probably a very unlikely edge case. It would have had to be deleted between the time the user opened edit mode, and clicked save.

I asked Maikel about different error codes from the backend and he said it was pretty complicated. So I like your idea of keeping it generic.

cooper-joe: OK, I think we can keep it generic with `Failed to save dashboard. You might be offline or not have access to edit this dashboard.`. The "might" is enough to suggest that these aren't the only possible reasons, but the most likely.

src/actions/dashboards.js
jenniferarnesen: Done

src/components/ControlBar/EditBar.js
jenniferarnesen: try/catch removed since it wasn't needed (not using async/await)

PR #1507, opened 2021-02-04
fix: restore star dashboard functionality

jenniferarnesen:
Fixed:
* the configuration for the mutation for api star dashboard was incorrect (id property not available for `create` mutations)
* display warning alert if the star dashboard or show description requests fail

Refactor/test
* Remove the thunk code and call the api function directly in the component, which can then handle request failures by showing an alert
* Cypress tests added for both success and fail scenarios

image

image


cooper-joe (approved):

src/api/starDashboard.js
```diff
@@ -0,0 +1,17 @@
+const getStarDashboardMutation = id => ({
+    type: 'create',
+    resource: `dashboards/${id}/favorite`,
+})
+
+const getUnstarDashboardMutation = id => ({
+    type: 'delete',
```
edoardo, line 7: Does the `delete` mutation also have the same problem as the `create` one when passing a function to `id`?

src/components/TitleBar/ViewTitleBar.js
```diff
@@ -64,19 +71,38 @@ const ViewTitleBar = props => {
...
+                warningAlert.show({ msg })
+            })
     }
 
     const toggleStarredDashboardLabel = starred
         ? i18n.t('Unstar dashboard')
         : i18n.t('Star dashboard')
 
-    const toggleStarredDashboard = () => {
-        // TODO use dataQuery directly here with create/delete mutations ?!
```
edoardo, line 74: My understanding of dataEngine usage is that if a query/mutation is only needed in one place, it should be as close as possible to where it's used.
Starring/unstarring only happens in this component and the mutation is not needed anywhere else, so I was wondering if it should be used here directly with the `useMutation` hook instead of define it in a separate file.
That should simplify the code a bit.
@amcgee can comment further here.

Some relevant discussions:
https://dhis2.slack.com/archives/G3TL0J145/p1602848442010400
https://dhis2.slack.com/archives/CLJSVA1RV/p1602144467062200
https://dhis2.slack.com/archives/CPGUFVC56/p1605263408418500

src/api/starDashboard.js
```diff
@@ -0,0 +1,17 @@
...
+    resource: `dashboards/${id}/favorite`,
+})
+
+export const apiStarDashboard = (dataEngine, id, isStarred) => {
+    const mutation = isStarred
+        ? getStarDashboardMutation(id)
```
edoardo, line 13: Do you get console errors complaining about the mutation not being static?
This is what @amcgee suggested for FileMenu:
https://github.com/dhis2/analytics/pull/655/commits/7a54cba875b1eee79640bbc679434ba0c930c7a8

src/components/TitleBar/ViewTitleBar.js
jenniferarnesen: That's fine with me.

One goal I have is to dynamically create Cypress api tests from the data query/mutations in the app. I haven't thought a lot about how to do that, but one thing I had in mind was to use convention by keeping all the query/mutation definitions in a standard location (e.g. `src/api`). But that shouldn't conflict with the idea of using the `useMutation` hook.

src/api/starDashboard.js
jenniferarnesen: No I'm not getting that warning.

I noticed the mutation in the link you posted had an `id` function: 

```
const mapMutation = {	
    resource: 'maps',	
    id: ({ id }) => id,	
    type: 'delete',	
}
```

Whereas star dashboard uses this: 

```
const getStarDashboardMutation = id => ({
    type: 'create',
    resource: `dashboards/${id}/favorite`,
})
```

Which doesn't have a function in it.  Not sure.

src/api/starDashboard.js
jenniferarnesen: I didn't check actually. I decided to go with the same style for star and unstar, and skip memoizing, which adds complexity.

src/components/TitleBar/ViewTitleBar.js
jenniferarnesen: Updated to use the `useDataMutation` hook.

src/api/starDashboard.js
jenniferarnesen: Update: when I switched to `useDataMutation`, I started seeing the warning, even when using `useMemo`.

src/components/TitleBar/ViewTitleBar.js
jenniferarnesen: Reverted to the original solution because of console warnings and failing cypress test

src/api/starDashboard.js
amcgee: `id` can be dynamic in a delete mutation

src/api/starDashboard.js
amcgee: The warnings are only when using declarative hooks, not when using imperative engine methods

src/components/TitleBar/ViewTitleBar.js
edoardo: 👍

edoardo (approved):

PR #1510, opened 2021-02-05
fix: height of interpretations panel in small screen [DHIS2-10425]

jenniferarnesen:
Background: the grid component we use expects item heights expressed as "grid units". The EXPANDED_HEIGHT and EXPANDED_HEIGHT_SM constants are expressed in grid units. That is the amount of height that is added to the item to make room for the interpretations container, which is hardcoded to (now) 334px.

Wide screen and small screen use different grid unit heights (this was to allow more space between items in small screen)

So the fix is to calculate the expanded height of the item correctly, using the correct grid units: 
Height of expanded area for wide screen: 17 grid units * 20px = 340px
Height of expanded area for small screen: 13 grid units * 26px = 338px

Meanwhile, the interpretations scroll container has a hardcoded height (now 334px), which will fit in the expanded area of the item for both wide and small screen.

Since a revamp of the interpretations functionality in dashboard is planned (probably using a modal instead of an expanding item), I decided not to refactor of this code (and getting rid of the hardcoded 334px height)

 <=480px

 greater than 480px


martinkrulltott (approved):
LGTM. Although would be interesting to know how this looks on a mobile device when switching from landscape to portrait, does it lag a bit like the More button did?

PR #1511, opened 2021-02-05
fix: add small More button for small screen [DHIS2-10424]

jenniferarnesen:
Issue was that the More button is together in a button strip containing other buttons. The other buttons are hidden in small screen, but More button is still available, albeit in "small".  This conversion to "small" was being handled by javascript. So there was a lag before the button changed. 

This proposed solution is to have two More buttons and they are hidden or shown using media query.

https://user-images.githubusercontent.com/6113918/107034280-c878c980-67b6-11eb-92ba-0db28f61a403.mov


src/components/TitleBar/ViewTitleBar.js
```diff
@@ -126,12 +123,21 @@ const ViewTitleBar = props => {
                             <FilterSelector />
                             <span ref={buttonRef}>
                                 <Button
-                                    small={isSmallScreen(width)}
-                                    dataTest="more-button"
+                                    className={classes.moreButton}
```
martinkrulltott: No biggie, but to reduce code repetition, parts of the button could be fetched from e.g. `const getMoreButton = className =>` or `const getMoreButtonContent = () =>`

src/components/TitleBar/styles/ViewTitleBar.module.css
```diff
@@ -55,10 +55,14 @@
     color: #888;
 }
 
-.moreButton {
+.moreText {
     margin-left: 5px;
 }
 
+.moreButtonSmall {
+    display: none !important;
```
martinkrulltott: Can `!important` be avoided by increasing the specificity with `.strip .moreButtonSmall`?

varl (commented):
Thanks for adding me, @jenniferarnesen. I think this use case may be common so I'll have it in mind if it can be made easier through UI.

src/components/TitleBar/styles/ViewTitleBar.module.css
jenniferarnesen: Done

src/components/TitleBar/ViewTitleBar.js
jenniferarnesen: Done

cooper-joe (approved):

martinkrulltott (approved):

PR #1514, opened 2021-02-05
fix: provide better message when edit mode becomes unavailable [DHIS2-10138]

jenniferarnesen:
image


martinkrulltott (approved):
`Resize your screen to return to edit`... is this really intuitive for mobile viewing?
How do I "resize" my phone?

cooper-joe (approved):
> Resize your screen to return to edit... is this really intuitive for mobile viewing?
How do I "resize" my phone?

My thinking behind this message is that this is as generic as we can be without ruling out other devices. "Resize your screen..." is not a perfect analogy for mobile, but it serves the purpose of covering all other devices like tablets in split screen, small desktop windows etc. It is not restrictive. 

Mentioning "landscape" or "turn your phone" suits the mobile device use case better, but in my opinion it is restrictive for other edge cases. Mentioning anything to do with a phone/mobile device will imply that the application cannot/should not be used on other devices, which isn't the case.

cooper-joe (changes requested):
I suggest using the previous message that does not reference landscape orientation or a mobile device.

cooper-joe (approved):

PR #1520, opened 2021-02-08
refactor: put system settings into context rather than redux

jenniferarnesen:
Redux is overkill for the system settings, since they do not change. Therefore, use React context instead. Note this may change again in the future if it becomes possible to get the settings from app-runtime

Notes:
* Visualization/Item - now consumes 2 contexts, so switched to using the `<ContextName.Consumer>` and away from `this.context`, which only allows for a single context.
* Visualization/Item test - removed complicated test that doesn't add any coverage over what Cypress tests provide
* reducers/settings (deleted file): please note all the properties that have been removed. They were not in use
* Removed file `storeDesign` - it was outdated, and not in use


jenniferarnesen:
@amcgee Could you comment on the part about using context for the system settings, and whether that might (ease/complicate) the migration to getting system settings from app-runtime sometime in the future?  I'll have Edoardo or Jan look at the rest of the PR.

src/api/settings.js
```diff
@@ -1,13 +1,11 @@
...
-export const apiFetchSystemSettings = () => {
-    const endPoint = '/systemSettings'
-    const url = `${endPoint}?${SYSTEM_SETTINGS.map(s => `key=${s}`).join('&')}`
-
-    return getInstance()
-        .then(d2 => d2.Api.getApi().get(url))
-        .catch(onError)
+const query = {
+    systemSettings: {
+        resource: `systemSettings?${SYSTEM_SETTINGS.map(s => `key=${s}`).join(
```
edoardo: I would prefer passing `params` like so:
`resource: 'systemSettings',
params: {
  key: SYSTEM_SETTINGS
  }`
In that way, `dataEngine` takes care of building the URL.

src/AppWrapper.js
```diff
@@ -48,9 +49,11 @@ const AppWrapper = () => {
                             return null
                         }
                         return (
-                            <WindowDimensionsProvider>
-                                <App />
-                            </WindowDimensionsProvider>
+                            <SystemSettingsProvider engine={dataEngine}>
```
edoardo: Can `dataEngine` be accessed directly in the Provider component?

src/components/App.js
```diff
@@ -27,16 +30,35 @@ import './App.css'
...
+    useEffect(() => {
+        const fetchDimensions = async () => {
+            try {
+                const dimensions = await apiFetchDimensions(
+                    dataEngine,
+                    settings.displayNameProperty
+                )
+
+                props.setDimensions(getFilteredDimensions(dimensions))
+            } catch (e) {
```
edoardo, line 52: Does this actually trigger if the fetch fails?

src/api/settings.js
```diff
@@ -1,13 +1,11 @@
...
 
-export const apiFetchSystemSettings = () => {
-    const endPoint = '/systemSettings'
-    const url = `${endPoint}?${SYSTEM_SETTINGS.map(s => `key=${s}`).join('&')}`
-
-    return getInstance()
-        .then(d2 => d2.Api.getApi().get(url))
-        .catch(onError)
+const query = {
+    systemSettings: {
```
edoardo: Wherever I used a query definition I ended up not putting the namespace in it.
In that way, whoever uses the query does not have to know how the namespace is called, but can decide what to use.
For example see:
https://github.com/dhis2/analytics/blob/ac78f355fc5712e1bddf0e90a390155731a9752a/src/api/dimensions.js#L8-L15
https://github.com/dhis2/analytics/blob/ac78f355fc5712e1bddf0e90a390155731a9752a/src/api/dimensions.js#L233-L239

src/components/SystemSettingsProvider.js
```diff
@@ -0,0 +1,42 @@
...
+}
+
+export const SystemSettingsCtx = createContext({})
+
+const SystemSettingsProvider = ({ engine, children }) => {
+    const [settings, setSettings] = useState([])
+
+    useEffect(() => {
+        async function fetchData() {
+            const { systemSettings } = await engine.query(settingsQuery)
```
edoardo: As mentioned above, I would rather use:
`const { systemSettings } = await engine.query({ systemSettings: settingsQuery })`

src/api/settings.js
jenniferarnesen: How to do it when the object key is the same for all of them? `key=keyname1&key=keyname2`

src/api/settings.js
jenniferarnesen: Updated

src/components/SystemSettingsProvider.js
jenniferarnesen: Updated

src/AppWrapper.js
jenniferarnesen: Yepp

src/components/App.js
jenniferarnesen: Yes, I checked both a 500 error and a 409 error. I haven't added any alerts for the user. I think I'll do that (and cypress test) in another refactor I want to do (wait to do the dimensions req until user opens filter dialog for first time)

src/api/settings.js
jenniferarnesen: Fixed

edoardo: Just pass the array of values, `dataEngine` is going to put them all together separated by `,`.
The backend supports both the multiple key and the single key with list of values formats.

src/components/App.js
edoardo: 👍

edoardo (approved):

PR #1521, opened 2021-02-09
fix: migrate inline height to css calculation for small screen control bar expanded, landscape and portrait

Sharmyn28:
Fixes in this PR:
- Small screen expanded control bar should take all the height -32px.
- Small screen expanded:
     - Add CSS variables for height calculation
     - Remove height calculation and inline-style
     - Control bar, dashboard bar, and chip container: Remove inline height use media query and CSS classes
- Small screen expanded (landscape):
     - Control bar expanded should take all the height -32px
     - Use media query and CSS classes to calculate the height
     - `Show more` button should be shown

**Small screen control bar expanded:**
image

**Small screen control bar expanded (landscape mode):**
image

**Small screen custom control bar height (landscape mode):**
image


jenniferarnesen (approved):

jenniferarnesen:
@Sharmyn28 Could you write up a good title and description for the PR, before merging it?

PR #1523, opened 2021-02-09
fix: make space to left of dashboard items scrollable [DHIS2-10138]

jenniferarnesen:
Fixes:
* use padding instead of margin so that scrolling works when scrolling the empty space along the left of the items
* migrate to consistent css classes language (wrapper => container)
* add buttons for both small and wide view and show/hide using css. This avoids the need for javascript to determine which button to show and will behave more instantaneously
* remove position:fixed on the action bar as it isn't needed (this was a relic from using the view dashboard as a template)

https://user-images.githubusercontent.com/6113918/107363291-389a8e80-6ada-11eb-96e0-ab25f0164fd0.mov

### DIFF
--- a/src/components/Dashboard/PrintActionsBar.js
+++ b/src/components/Dashboard/PrintActionsBar.js
@@ -4,8 +4,6 @@ import i18n from '@dhis2/d2-i18n'
 import { Button } from '@dhis2/ui'
 import { Link } from 'react-router-dom'
 import LessHorizontalIcon from '../../icons/LessHorizontal'
-import { useWindowDimensions } from '../WindowDimensionsProvider'
-import { isSmallScreen } from '../../modules/smallScreen'
 
 import classes from './styles/PrintActionsBar.module.css'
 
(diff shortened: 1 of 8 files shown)


src/components/Dashboard/PrintActionsBar.js
```diff
@@ -14,25 +12,41 @@ export const PRINT_ACTIONS_BAR_HEIGHT = 44
 export const PRINT_ACTIONS_BAR_HEIGHT_SM = 36
 
 const PrintActionsBar = ({ id }) => {
-    const { width } = useWindowDimensions()
-    const isSmall = isSmallScreen(width)
+    const getExitPrintButton = isSmall => {
+        const className = isSmall ? classes.buttonSmall : classes.buttonWide
```
martinkrulltott: I'm not sure what the best naming convention is, but could we possibly use `small - large` or maybe `narrow - wide`? `small - wide` sound like two different properties to me (however your English is better than mine so I could be wrong).

martinkrulltott: I'd also consider moving the ternary operator directly into the `className={ ... }` to reduce the footprint a bit.

src/components/Dashboard/styles/PrintDashboard.module.css
```diff
@@ -14,12 +13,17 @@
...
+    .container {
+        padding-left: var(--spacers-dp12);
+    }
+}
+
 @media print {
-    .wrapper {
+    .container {
         background-color: white;
         height: auto !important;
```
martinkrulltott, line 25: Probably out of scope, but I noticed these `!important`. `media -> class` should override any `class` selector, so are these really necessary?

src/components/Dashboard/styles/PrintDashboard.module.css
```diff
@@ -14,12 +13,17 @@
...
         overflow: visible !important;
-        margin-left: 0px;
         padding-left: 0px;
```
martinkrulltott, line 27: Also not in scope, and it has no production code impact, but my personal preference is to use `0` instead of `0px`. This is just my OCD speaking, so disregard this if you have a different opinion 😄

src/components/Dashboard/styles/PrintLayoutDashboard.module.css
```diff
@@ -1,7 +1,6 @@
-.wrapper {
+.container {
     background-color: #f4f6f8;
```
martinkrulltott, line 2: Yet another scope-creeping comment, but should this color be fetched from `ui`? Ok, I'm sorry, I'll stop now 😉

martinkrulltott (approved):
LGTM 🏅  the actual change was so good I had to move out of scope to find things to comment 😆   Anyway, it's good to merge as is, just couldn't help myself to write comments while I was reviewing 🍻

src/components/Dashboard/styles/PrintDashboard.module.css
jenniferarnesen: I've been starting to review a lot of the css and I'll make a note to look at this one. Thanks!

src/components/Dashboard/styles/PrintLayoutDashboard.module.css
jenniferarnesen: This color isn't in ui. I've checked with Joe on this before. Will check again if there are plans for it to go into the color suite.

PR #1540, opened 2021-02-15
feat: restrict view options of dashboard items [DHIS2-7630]

jenniferarnesen:


jenniferarnesen:
Not needed after all.

PR #1546, opened 2021-02-17
feat: lazily load plugin scripts and dependencies when needed (DHIS2-10518)

amcgee:
1. Remove blocking external javascript script tags from index.html head
2. Dynamically load legacy plugin scripts when they are needed, not before
3. Dynamically `import()` the Data Visualizer plugin so that it can be code-split out of the main bundle. (NOTE: for this to really be useful it requires a tree-shakable analytics library, which we don't have yet)

This also removes the need for a custom index.html and for setting any environment variables in development

DV and Maps plugins will hopefully move to dynamic modules, but this solution will still be required for eventReports and eventVisualizer

Here is a production example : https://debug.dhis2.org/2.35dev/api/apps/dashboard2/index.html#/Goz4vyRx2cy

NOTE: ER and EV plugins don't work locally because of CORS

TODO:
- [x] Decide whether to use `prefetch`, `preload`, or neither for `data-visualizer-plugin` chunk (currently prefetching)
- [x] Investigate EV EXT error
- [x] Show `<LoadingMask>` while fetching scripts
- [x] Remove console.logs


jenniferarnesen:
- [ ]  only map - loads map.js and dv plugin. Dashboard without map - no map.js
- [ ]  no visualizations should not load any plugins. Actual - loads dv plugin, but not the other plugins
- [ ] only ev - loads eventchart.js and dv plugin chunk
- [ ] only er - loads eventreport.js and dv plugin chunk
- [x] only chart/pivot - loads dv plugin chunk and no other plugins

I'm getting a bunch of errors on the production instance resulting in ev not rendering:
```
3UiManager.js?d38e:536 Uncaught ReferenceError: Ext is not defined
    at m.n.alert (UiManager.js?d38e:536)
    at Object.<anonymous> (Layout.js?bf56:459)
    at u (jquery-3.3.1.min.js:2)
    at Object.fireWith [as rejectWith] (jquery-3.3.1.min.js:2)
    at k (jquery-3.3.1.min.js:2)
    at XMLHttpRequest.<anonymous> (jquery-3.3.1.min.js:2)
```
Dashboard with only a text item
image

Dashboard with a chart and pivot table
image

amcgee:
@jenniferarnesen the duplication of `data-visualizer-plugin` chunk there is expected - the `javascript` type is from [a `<link>` element with `rel="prefetch"`](https://www.w3schools.com/tags/att_link_rel.asp) inject into the html by webpack to prefetch and cache the script as early as possible, but not block rendering.  That way it's immediately available if it's needed later.  I'd like to investigate whether we want to switch to `preload` instead... but I don't think this shouold be detrimental here.

> prefetch | Specifies that the browser should preemptively fetch and cache the target resource as it is likely to be required for a follow-up navigation
> preload | Specifies that the browser agent must preemptively fetch and cache the target resource for current navigation according to the destination given by the "as" attribute (and the priority associated with that destination).

amcgee:
Strange that we're seeing the EV EXT error now though.... i am seeing it too but it was working for me before.  Will have to investigate that.

amcgee:
Re: the EV EXT error

This error also appears on the current dashboard https://debug.dhis2.org/2.35dev/dhis-web-dashboard/#/PZIxO0YLxFt -- the eventchart.js plugin tries to use EXT to show an alert saying "the user admin doesn't have access to this program", but there's no EXT available (that's only in the app itself).  It works with the system user, which has access to all programs.  So this appears to be a bug in the EV plugin and/or we need to also load EXT when loading that plugin (which can be done but... ew)

jenniferarnesen (approved):

amcgee:
@jenniferarnesen I've updated this a bit (better late than never) to add loading mask and remove console logs, the implementation changed a bit to support this.  Think we can get this in to 2.36 before the hard freeze?

src/components/Item/VisualizationItem/Visualization/Visualization.js
```diff
@@ -84,29 +83,21 @@ class Visualization extends React.Component {
...
+                                pluginProps.itemFilters
                             )}
-                        </D2Shim>
+                            onLoadingComplete={this.onLoadingComplete}
+                            forDashboard={true}
+                            style={pluginProps.style}
+                            userSettings={{
+                                displayProperty: this.context.userSettings
+                                    .keyAnalysisDisplayProperty,
+                            }}
```
jenniferarnesen: `forDashboard`, `userSettings` and `onLoadingComplete` could be moved to DataVisualizerPlugin. Should be pretty straightforward (onLoadingComplete just sets a state var that is only relevant for DataVisualizerPlugin).  But if you don't have time for this, I can do it. If its not done by noon (either by me or you), then let's just go ahead and merge as is, and I can do this later.

jenniferarnesen (approved):

src/components/Item/VisualizationItem/Visualization/Visualization.js
jenniferarnesen: Changes committed

amcgee:
@jenniferarnesen nice, much cleaner thanks!  Changes look good, shall we merge?  👍

PR #1550, opened 2021-02-18
feat: scroll away control bar when viewing dashboard on screen with height <=480px

jenniferarnesen:
When the screen height is <=480px, then scroll all the content, including the control bar. When more than 480px, scroll only the actual dashboard. Note that progressive loading is not working consistently right now (pre-existing issue - currently being investigated)

Small refactor:
* move styling from App.css and into a css module on a new component DashboardContainer

https://user-images.githubusercontent.com/6113918/108341782-5488e900-71da-11eb-98c2-9ea04bab72d2.mov

### DIFF
--- a/src/components/Dashboard/ViewDashboard.js
+++ b/src/components/Dashboard/ViewDashboard.js
@@ -29,7 +29,7 @@ export const ViewDashboard = props => {
 
     useEffect(() => {
         Array.from(
-            document.querySelectorAll('.dashboard-scroll-container')
+            document.getElementsByClassName('dashboard-scroll-container')
         ).forEach(container => {
             container.scroll(0, 0)
         })
--- a/src/components/Dashboard/ViewDashboard.js
+++ b/src/components/Dashboard/ViewDashboard.js
@@ -3,6 +3,7 @@ import { connect } from 'react-redux'
 import PropTypes from 'prop-types'
 import { ComponentCover } from '@dhis2/ui'
 import cx from 'classnames'
+import DashboardContainer from './DashboardContainer'
 import ViewTitleBar from '../TitleBar/ViewTitleBar'
 import ViewItemGrid from '../ItemGrid/ViewItemGrid'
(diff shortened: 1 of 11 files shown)


src/components/Dashboard/ViewDashboard.js
```diff
@@ -27,24 +28,22 @@ export const ViewDashboard = props => {
     }, [props.dashboardIsEditing, props.dashboardIsPrinting])
 
     useEffect(() => {
-        document.querySelector('.dashboard-wrapper')?.scroll(0, 0)
+        Array.from(
+            document.getElementsByClassName('dashboard-scroll-container')
+        ).forEach(container => {
+            container.scroll(0, 0)
+        })
```
edoardo, line 35: Why not using `querySelector` still?
I understand there is always only 1 element with that class.
Or maybe use this instead?
```suggestion
            document.getElementsByClassName('dashboard-scroll-container').item(0)?.scroll(0,0)
```

src/components/Item/ProgressiveLoadingContainer.js
```diff
@@ -51,23 +51,27 @@ class ProgressiveLoadingContainer extends Component {
             this.props.debounceMs
         )
 
-        document
-            .getElementsByClassName('dashboard-wrapper')[0]
-            ?.addEventListener(
+        Array.from(
+            document.getElementsByClassName('dashboard-scroll-container')
+        ).forEach(container => {
+            container.addEventListener(
```
edoardo, line 57: See comment above.

edoardo (approved):

src/components/Dashboard/ViewDashboard.js
jenniferarnesen: Actually, there can be 2 elements with the class

edoardo: Yes, i missed the one in `DashboardContainer`.
Nvm the comments.

PR #1554, opened 2021-02-19
fix: debounce the item selector query in dashboard editor [DHIS2-10530]

jenniferarnesen:
Fix:
- prevent the item selector query from firing on every keystroke.

Refactor:
- useDataEngine instead of manually constructing the query
- SinglesMenuGroup exports was triggering eslint warning, so no longer export const SinglesMenuGroup (unit test had to be updated)
- remove USERS from ItemSelector as it is no longer meant to be used (confirmed with @janhenrikoverland )
- borrow the `useDebounce` hook from analytics. Have suggested to Austin that it could be added to app-runtime

### DIFF
--- a/src/components/ItemSelector/ItemSelector.js
+++ b/src/components/ItemSelector/ItemSelector.js
@@ -1,11 +1,13 @@
 import React, { useState, useEffect, createRef } from 'react'
 import { Popover, FlyoutMenu } from '@dhis2/ui'
-import { useD2 } from '@dhis2/app-runtime-adapter-d2'
+import { useDataEngine } from '@dhis2/app-runtime'
 import ItemSearchField from './ItemSearchField'
 import CategorizedMenuGroup from './CategorizedMenuGroup'
 import SinglesMenuGroup from './SinglesMenuGroup'
 import { singleItems, categorizedItems } from './selectableItems'
 import { itemTypeMap, getDefaultItemCount } from '../../modules/itemTypes'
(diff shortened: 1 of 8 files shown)


src/components/ItemSelector/ItemSelector.js
```diff
@@ -14,21 +16,17 @@ const ItemSelector = () => {
     const [filter, setFilter] = useState('')
     const [items, setItems] = useState(null)
     const [maxOptions, setMaxOptions] = useState(new Set())
-    const { d2 } = useD2({})
+    const dataEngine = useDataEngine()
+    const debouncedFilterText = useDebounce(filter, 200)
```
janhenrikoverland: I think we should set 500 to get more out of it @jenniferarnesen @cooper-joe

janhenrikoverland (approved):
Approved when comment is resolved.

src/components/ItemSelector/ItemSelector.js
cooper-joe: I would argue that 500ms is too long. For form validation the debounce can be longer, but for search the user is expecting to see results, so any delay is purely technical in nature. I would say 350ms should be the maximum, unless there's a strong technical backing to delay it further. (Related: [Doherty Threshold](https://lawsofux.com/doherty-threshold/))

janhenrikoverland: Good point, @cooper-joe. Let's go with 350 ms.

PR #1562, opened 2021-02-22
feat: register passive view of dashboards [DHIS2-7016]

tomzemp:
* creates actions/reducers for tracking if passive view has been registered
* posts initial view to dataStatistics from ViewDashboard as passive view


tomzemp:
This implementation posts to data statistics when the dashboard app renders a dashboard in view mode for the first time (a passive view); any subsequent passive views, regardless of dashboard, are not registered. In line with the back-end implementation by @jimgrace, the dashboard app has been modified to always make these posts for passive views, whereas api/dataStatistics will count these passive views only if the newly added system setting keyCountPassiveDashboardViewsInUsageAnalytics=true.

This change is requested by PEPFAR to allow better oversight over which dashboards are being accessed by users. This change will not affect statistics unless a system admin chooses to set keyCountPassiveDashboardViewsInUsageAnalytics=true, so this PR does not change default behaviour for existing instances (https://jira.dhis2.org/browse/DHIS2-7016)

jenniferarnesen (approved):

jenniferarnesen:
Duplicated in another PR for the PR checks

PR #1563, opened 2021-02-22
fix: controlbar should always take full width, and other minor position adjustments

jenniferarnesen:
Controlbar wasn't taking full width when searching:
image

After:
image

Get rid of scrollbar when chips don't take full width:
Before: 
image
After:

image

Position of search icon was too low
Before: 
image

After: 
image


janhenrikoverland (approved):
LG

PR #1566, opened 2021-02-24
fix: get value for margins and padding from css when calculating item available height

jenniferarnesen:
Previously the values margins and padding needed for calculating item height were stored in duplicate (value in js and in css). This was error prone since one value could be changed leading to a mismatch. Now, the javascript gets the value from the stylesheet.

Before:
image

After fix:

image


turban (approved):
LGTM!

PR #1567, opened 2021-02-24
fix: dashboard filter modal should not be available in small screen

jenniferarnesen:
Adds a separate element in the FilterBadge for small screen that does not have the click action, so that the filter modal cannot be opened while in small screen. (video not included, but the cypress tests covers the scenario)

image

2 new cypress tests added - one where it is possible to open the filter modal, and one where it isn't possible.

Jest unit test added to check that Filter badge has the correct text on it.


src/components/FilterBar/__tests__/FilterBadge.spec.js
```diff
@@ -0,0 +1,44 @@
...
+import configureMockStore from 'redux-mock-store'
+import FilterBadge from '../FilterBadge'
+
+const mockStore = configureMockStore()
+
+test('Filter Badge displays badge containing number of items in filter', () => {
+    const filter = {
+        id: 'ponies',
+        name: 'Ponies',
+        values: [{ name: 'Rainbow Dash' }, { name: 'Twilight Sparkle' }],
```
martinkrulltott, line 13: Now I finally know who added all of the my little pony names to DV 🦄 😄

martinkrulltott (approved):

src/components/FilterBar/__tests__/FilterBadge.spec.js
jenniferarnesen: 🦄

PR #1593, opened 2021-03-02
fix: do not allow "Open in [appname] app" when in small screen

jenniferarnesen:
It has been decided that when in small screen, since the Data Visualizer, Maps, EV, ER apps do not work well on small screen, don't give the option to navigate to those apps from the dashboard when in small screen.

This should be merged to branch v36.

image

Refactor:
- switch from enzyme to testing-library and add more tests to cover the somewhat messy logic
- rename from ItemHeaderButtons to ItemContextMenu
- break out the ViewAsMenuItems component from ItemContextMenu to separate some of the complexity (no logic changed in ViewAsMenuItems)
- give consistent names to the system settings
- fix a few test warnings in unrelated tests (ViewDashboard, EditTitleBar)


src/components/Item/VisualizationItem/ItemContextMenu/ViewAsMenuItems.js
```diff
@@ -0,0 +1,67 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+import i18n from '@dhis2/d2-i18n'
+import TableIcon from '@material-ui/icons/ViewList'
+import ChartIcon from '@material-ui/icons/InsertChart'
+import MapIcon from '@material-ui/icons/Public'
```
edoardo, line 6: Can these be easily replaced with `ui-icons`?

edoardo (approved):
Looks good.
Perhaps replace MUI icons with `ui-icons`?

src/components/Item/VisualizationItem/ItemContextMenu/ViewAsMenuItems.js
jenniferarnesen: I'm going to do all the icons in a separate branch - didn't want to make this PR too big :)

PR #1596, opened 2021-03-03
fix: decrease dashboard request response load size

jenniferarnesen:
Request load improvements:
* Limit the request that fetches all dashboards (for the dashboard bar) to only getting `id, displayName, favorite~rename(starred)`. This reduces the response size from 3.9kb to 1.1kb on the SL database. Feels like a drop in the bucket, but that will have a bigger impact on instances with a lot of dashboards
* Create separate requests getting a single dashboard for viewing and editing. In view, there are several fields that aren't used. In edit, we need all the fields if we are going to PUT changes.

Refactor:
* move getCustomDashboards to own module file
* split up api requests for dashboards into separate files


janhenrikoverland (approved):
LGTM

PR #1600, opened 2021-03-04
fix: use keyAnalysisDisplayProperty from user settings when requesting dimensions for filter

jenniferarnesen:
image

Refactor:
* remove hardcoded displayNameProperty that was together with system settings (not where it belonged either)
* remove other userSettings request and use the useUserSettings hook instead
* Messages Item - move style to css module


martinkrulltott (commented):
`userSettings.displayProperty `needs to be passed to the plugin.
And `displayNameProp` to DynamicDimension.

jenniferarnesen:
> `userSettings.displayProperty `needs to be passed to the plugin.
> And `displayNameProp` to DynamicDimension.

Updated

martinkrulltott (approved):

PR #1620, opened 2021-03-09
fix: use window height to determine expanded control bar height on small screens

jenniferarnesen:
This will hopefully fix the issue where the expanded control bar on the phone is too tall. Using window height should take into account the phone's available screen space considering address bar and navigation controls on some phone models.

image


src/components/App.css
```diff
@@ -12,6 +12,9 @@
     --user-rows-height: calc(
         var(--controlbar-padding) + (var(--user-rows-count) * var(--row-height))
     );
+    --sm-expanded-controlbar-height: calc(
+        (var(--vh, 1vh) * 100) - var(--headerbar-height) - 32px
```
martinkrulltott, line 16: why `32px`?

martinkrulltott (approved):

PR #1624, opened 2021-03-09
fix: fix double border in More menu

edoardo:
Before:
Screenshot 2021-03-09 at 16 13 34

After:
Screenshot 2021-03-09 at 16 13 01


jenniferarnesen (approved):

edoardo:
@cooper-joe it doesn't seem to be possible to have the arrow but not the double border with the current `ui` components.
`Popover` has the arrow feature, but when used in combination with `FlyoutMenu` has the double border issue.
`Menu` (which is not wrapped in `Card`) would work in that scenario, but the sublevels don't open (Print submenu).
What should we do?

cooper-joe:
> @cooper-joe it doesn't seem to be possible to have the arrow but not the double border with the current ui components.

The arrow is not required or recommended, so this is not a problem.

PR #1627, opened 2021-03-10
fix: catch errors from the interpretations component

jenniferarnesen:
Two fixes:
* catch interpretations crashes
* when an item crashes, only show "Open in App" option in the context menu

image

https://user-images.githubusercontent.com/6113918/110661785-2e48de80-81c5-11eb-9763-e16da3b17973.mov

https://user-images.githubusercontent.com/6113918/110661793-3143cf00-81c5-11eb-8802-acc1d5ce8866.mov


jenniferarnesen:
@cooper-joe : by removing the item context menu, the user can't click on "View in Visualizer app". Is that acceptable? An alternative would be to still have the menu with just that button.

cooper-joe:
> by removing the item context menu, the user can't click on "View in Visualizer app". Is that acceptable? An alternative would be to still have the menu with just that button.

I think "View in Visualizer..." is quite a useful feature for errored items as it may help the user identify what the error is? I think keeping the menu with that button would be ideal. In the future, we could also add a "Reload" or "Retry loading this item" in that menu, so I think it's useful to keep.

edoardo (approved):

jenniferarnesen:
> > by removing the item context menu, the user can't click on "View in Visualizer app". Is that acceptable? An alternative would be to still have the menu with just that button.
> 
> I think "View in Visualizer..." is quite a useful feature for errored items as it may help the user identify what the error is? I think keeping the menu with that button would be ideal. In the future, we could also add a "Reload" or "Retry loading this item" in that menu, so I think it's useful to keep.

@cooper-joe Updated - see screenshot in description

cooper-joe (approved):

PR #1632, opened 2021-03-11
fix: disable View As menu items for maps without a thematic layer [DHIS2-10671]

jenniferarnesen:
If a map doesn't have a thematic layer, then no chart or table can be rendered. Therefore, disable these menu items and add a tooltip informing the user

image


turban (approved):
LGTM!

PR #1634, opened 2021-03-11
fix: upgrade to @dhis2/cli-app-scripts@6 (DHIS2-9893)

amcgee:
Part of [DHIS2-9893](https://jira.dhis2.org/browse/DHIS2-9893)

Upgrades ui, cli-app-scripts, app-runtime, d2-i18n, and analytics

Tree shaking combined with lazy plugin loading gives us a savings of 1.5MB (!!!!) javascript payload (gzipped) when no EV or ER plugins are present.  Those add another ~1MB though.

Maps plugin slimming will increase the benefit here as well.

Before
Screen Shot 2021-03-11 at 19 10 48

After
Screen Shot 2021-03-11 at 19 10 41

---

The initial render-blocking load (before plugin code is fetched) is reduced from 3.1MB gzipped 🥴   to just 547kb 🎉  -- still too much and we should look into non-js bloat but MUCH improved!

Before
Screen Shot 2021-03-11 at 19 36 30

After
Screen Shot 2021-03-11 at 19 35 58

### DIFF
--- a/d2.config.js
+++ b/d2.config.js
@@ -1,11 +1,11 @@
 const config = {
-    name: 'dashboard',
     type: 'app',
-    coreApp: true,
+    name: 'dashboard',
     title: 'Dashboard',
+    coreApp: true,
 
     entryPoints: {
-        app: './src/AppWrapper',
+        app: './src/AppWrapper.js',
     },
 }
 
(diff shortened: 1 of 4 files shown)


varl:
Pretty slick for what is basically a two file update!

varl (approved):
Tested locally and seems to work fine to me.

PR #1658, opened 2021-03-19
fix: avoid double rendering of chart visualizations

edoardo:
Requires: https://github.com/dhis2/data-visualizer-app/pull/1665

A few issues caused the double rendering of chart visualizations, which causes all the code for generating Highcharts config to run twice.
This is a problem especially for some chart types and options (scatter with outliers detection) which needs expensive computation.

* 2 props passed to `VisualizationPlugin` were not the same on subsequent rendering of the parent component, causing unwanted rendering and triggering effects
* the `id` prop is not passed to `VisualizationPlugin` and assumed `undefined` value, triggering a re-render in an effect which checks if `id` is different than `null` (the fix for this is in DV plugin, see PR linked above)

### DIFF
--- a/src/components/UserSettingsProvider.js
+++ b/src/components/UserSettingsProvider.js
@@ -15,7 +15,10 @@ const UserSettingsProvider = ({ children }) => {
                 userSettings: userSettingsQuery,
             })
 
-            setSettings(userSettings)
+            setSettings({
+                ...userSettings,
+                displayProperty: userSettings.keyAnalysisDisplayProperty,
+            })
         }
         fetchData()
     }, [])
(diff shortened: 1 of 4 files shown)


src/components/UserSettingsProvider.js
```diff
@@ -15,7 +15,10 @@ const UserSettingsProvider = ({ children }) => {
                 userSettings: userSettingsQuery,
             })
 
-            setSettings(userSettings)
+            setSettings({
+                ...userSettings,
+                displayProperty: userSettings.keyAnalysisDisplayProperty,
```
jenniferarnesen, line 20: If there is time, I would prefer to not duplicate this property in the result (now it has both `displayProperty` and `keyAnalysisDisplayProperty`).  You can see it being used in FilterDialog.  One approach could be to give better names for all the returned settings (similar is done in the system settings (api/systemSettings and SystemSettingsProvider). If there isn't time for this, we can do it later.  So for instance keyDbLocale->dbLocale, keyUiLocale->uiLocale.

martinkrulltott (approved):

jenniferarnesen (approved):

PR #1659, opened 2021-03-19
fix: maps, ev and er were not resizing for fullscreen or item resize in edit mode

jenniferarnesen:
Since the switch to dynamic module loading, the plugins are now promises and need to be resolved. There were some `await`s that were missing.

Other improvements were to move the listener for resizing events to the specific plugin component rather than having it in componentDidMount in the generic Visualization component. This reduced the number of times re-render of the plugin was triggered.

Fullscreen:

https://user-images.githubusercontent.com/6113918/111987152-2109e900-8b0f-11eb-9d42-26b384567c83.mov

Window resize:

https://user-images.githubusercontent.com/6113918/111987192-2b2be780-8b0f-11eb-8779-826b225568b9.mov

Item resize:

https://user-images.githubusercontent.com/6113918/111987212-33842280-8b0f-11eb-8574-3d2dc6f9d5cb.mov

### DIFF
--- a/package.json
+++ b/package.json
@@ -15,7 +15,7 @@
         "@dhis2/d2-ui-rich-text": "^7.1.6",
         "@dhis2/d2-ui-sharing-dialog": "^7.1.6",
         "@dhis2/d2-ui-translation-dialog": "^7.1.6",
-        "@dhis2/data-visualizer-plugin": "^35.20.19",
+        "@dhis2/data-visualizer-plugin": "^35.20.20",
         "@dhis2/ui": "^6.5.5",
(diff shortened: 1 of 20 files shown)


turban:
@jenniferarnesen, can you confirm that scroll wheel zoom is disabled for map items by default, but enabled when in fullscreen mode?

jenniferarnesen:
> @jenniferarnesen, can you confirm that scroll wheel zoom is disabled for map items by default, but enabled when in fullscreen mode?

Yes that describes current behavior

turban (approved):
LGTM. Will do some extra testing after this is merged to master.

PR #1662, opened 2021-03-23
chore: complete restructure of files and folders in src

jenniferarnesen:
Yes there is a massive number of files in this PR, but the majority of them have either slight (change to dependency path) or no changes.

Organize components into "pages" loosely based on the available routes:
* view
* edit (both edit and new)
* print (both print layout and print one-item-per-page).

Any components that are used by more than one page remain in the shared components (src/components) folder.  Any modules or api's that are only used by one page are moved to that page folder.

Removed from redux since they were only used in one place
* alert (and corresponding Alert components)
* user

components/Dashboard is nearly the same as it was previously as components/Dashboard/Dashboard. The only change is to handle the show/hiding of the Alert when a dashboard is taking a long time to load (used to be in actions/selected)

ListItem -> move some functionality from thunk directly into component since it seemed overengineered

getFormattedDate taken out of util.js and into its own module file. Also removed unused functions from util.js

Added an alert when deleting dashboard fails

grid css (3 files) is now imported in components/App.js - which is the root of the app. Before it was being imported by each of the grid components (edit, view, print).

removed other unused actions, and unused widgets


janhenrikoverland (approved):
LG, let's see how it works out

PR #1663, opened 2021-03-23
fix: ctrl bar covered sharing dialog and cypress tests [DHIS2-10691] [v36]

jenniferarnesen:
The z-index was causing the ctrl bar to cover the sharing dialog, which overlay has a lower z-index. The z-index on the control bar is only needed when it is expanded on small screen, in which case the Sharing dialog is not an option anyway.

Cypress test added.


jenniferarnesen:
duplicate

PR #1666, opened 2021-03-23
fix: maps, ev and er were not resizing for fullscreen or item resize in edit mode [v36]

jenniferarnesen:
Since the switch to dynamic module loading, the plugins are now promises and need to be resolved. There were some awaits that were missing.

Other improvements were to move the listener for resizing events to the specific plugin component rather than having it in componentDidMount in the generic Visualization component. This reduced the number of times re-render of the plugin was triggered.

Link to PR that was merged to master (contains screenshots or video): https://github.com/dhis2/dashboard-app/pull/1659


turban (approved):
LGTM

PR #1672, opened 2021-03-24
fix: check dom element directly to determine if element is fullscreen in order to avoid unnecessary visualization rerenders

jenniferarnesen:
Instead of setting a state to trigger changes related to going fullscreen, the components that need to observe the state of fullscreen will now use the native web api and dom elements. This avoids a second render that was happening when going fullscreen.

The onResizeStop listener isn't needed anymore since each plugin individually handles size changes (in MapPlugin, LegacyPlugin, DV Plugin)

Some of the files are just function renaming.

Other improvement: do not create an ItemContextMenu when in edit mode.


src/components/Item/VisualizationItem/ItemContextMenu/__tests__/ItemContextMenu.spec.js
```diff
@@ -47,17 +47,28 @@ test('renders just the button when menu closed', () => {
 })
 
 test('renders exit fullscreen button', () => {
-    useSystemSettings.mockImplementationOnce(() => mockSystemSettingsDefault)
-    const props = Object.assign({}, defaultProps, {
-        isFullscreen: true,
-    })
-    const { container } = render(
+    useSystemSettings.mockImplementation(() => mockSystemSettingsDefault)
+    const gridItemClassName = `reactgriditem-${defaultProps.item.id}`
```
martinkrulltott:
```suggestion
    const gridItemClassName = getGridItemDomElementClassName(defaultProps.item.id)
```

martinkrulltott: Never mind, didn't realise this was the test code 🙈

jenniferarnesen: Did it anyway.

martinkrulltott (approved):

PR #1678, opened 2021-03-26
fix: makes Add filter button display only if there are available dimensions to filter

jenniferarnesen:
* hides Add filter button in dashboard control bar if no allowed filters
* adds related unit tests


jenniferarnesen:
Already merged in Tom's PR

PR #1692, opened 2021-03-31
fix: bump dv plugin to ^34.6.57 for v34

martinkrulltott:


martinkrulltott:
Note that the commit message is wrong, should've been `34.6.57`

jenniferarnesen (approved):

PR #1700, opened 2021-04-12
feat: offline dashboard

jenniferarnesen:
Implements DHIS2-10874

Add offline functionality to dashboard to support Dashboard PWA.  With this implementation, it is possible to cache dashboards so they can be viewed while offline. The app also detects the offline/online state, and updates availability of actions - like editing, starring, changing filters, and viewing certain info like sharing and interpretations -  accordingly.

Much of this PR is about disabling/enabling buttons and menu items, and showing a tooltip. ViewDashboard is the main code for navigating to different dashboards while offline, while view/TitleBar/ActionsBar triggers the recording of a dashboard.

Additional info for some of the changed files: 

 Cypress tests:
- all new tests related to offline are in the `offline.feature` and `offline.js` files. The rest of the changes are adaptations based on changed html structure, or sharing functionality
- `goOffline` and `goOnline` functions were copied from this post: https://www.cypress.io/blog/2020/11/12/testing-application-in-offline-network-mode/

d2.config.js
- this is the config that enables offline functionality in the app-runtime. The `patternsToOmit` property is a list of endpoints that should NOT be cached as part of the GENERAL app cache. Therefore, the endpoints listed in this property are related to content for a specific dashboard. Endpoints that go into the general app cache include e.g. `/dashboards`, static assets (dashboard bundle, plugins, jquery...), `/schemas`, etc.  

actions/selected.js
- refactored `tSetSelectedDashboardById` to use async/await
- `tSetSelectedDashboardByIdOffline` - this is used when user navigates to an uncached dashboard while offline. We still want to navigate to the url and set the state of the app, then when connection is restored, the rest of the data gets loaded.

Tooltip.js
- this component is used throughout the app to apply a tooltip when the button is disabled.
MenuItemWithTooltip.js
- applies a tooltip to MenuItems - used for offline messages.

ConfirmActionDialog.js (refactor/code sharing)
- used by 3 other components

DropdownButton.js
- modifed to handle additional property `disabledWhenOffline` - moved that and other properties into `rest`
-

VisualizationItem/Item.js
-  need to always fetch visualization in ComponentDidMount, to handle case of dashboard being recorded.

ViewAsMenuItems.js
- uses MenuItemWithTooltip, and the offline status will override the supplied tooltip.

DefaultPlugin.js
-  unmount visualization when the item is being removed from page. TODO: confirm changes with Jan.

ProgressiveLoadingContainer.js
- `forceLoadCount` is an incrementing property that triggers force loading when a dashboard is being recorded.

PrintDashboard, PrintLayoutDashboard
- bug fixes for bugs that the offline functionality revealed.

CacheableViewDashboard
- adds `CacheableSection` which enables the dashboard to be recorded as a section.

DashboardsBar - useCallback to prevent unneeded rerendering of the content

FilterBadge.js, FilterBar.js
- only allow removing filter if online, or if cached then user can remove, but has to remove all filters

view/TitleBar.js
- incorporate the Last updated tag

ViewDashboard.js
- complex logic in useEffect to determine how to handle opening dashboards depending on the offline/online state and whether the dashboard is cached. The cypress tests cover most of the scenarios.

yarn.lock.
- yeah, just don't.  I tried merging in changes from master but there were so many conflicts that were really difficult to resolve, that I ended up deleting the file and regenerating it.  So lots of dependencies that have nothing to do with this feature are unfortunately in there.


src/pages/view/TitleBar/LastUpdatedTag.js
```diff
@@ -0,0 +1,27 @@
...
+    const { lastUpdated } = useCacheableSection(id)
+
+    return lastUpdated && lastUpdated.toString ? (
+        <Tooltip
+            content={<span>{moment(lastUpdated).format('llll')}</span>}
+            openDelay={200}
+            closeDelay={100}
+        >
+            <Tag className={classes.lastUpdatedTag} maxWidth="400px">
+                {`Offline data last updated ${moment(lastUpdated).fromNow()}`}
```
martinkrulltott: i18n?

src/pages/view/TitleBar/LastUpdatedTag.js
```diff
@@ -0,0 +1,27 @@
...
+import { Tag, Tooltip } from '@dhis2/ui'
+import moment from 'moment'
+import PropTypes from 'prop-types'
+import React from 'react'
+import classes from './styles/LastUpdatedTag.module.css'
+
+const LastUpdatedTag = ({ id }) => {
+    const { lastUpdated } = useCacheableSection(id)
+
+    return lastUpdated && lastUpdated.toString ? (
```
martinkrulltott:
```suggestion
    return lastUpdated?.toString ? (
```

src/pages/view/FilterBar/styles/FilterBadge.module.css
```diff
@@ -22,12 +22,29 @@
 }
 
 .removeButton {
+    background-color: transparent;
+    color: white;
```
martinkrulltott:
```suggestion
    color: var(--colors-white);
```

src/pages/view/FilterBar/styles/FilterBadge.module.css
```diff
@@ -22,12 +22,29 @@
...
     margin-left: var(--spacers-dp24);
     cursor: pointer;
 }
 
+.span {
+    display: inline-flex;
+    pointer-events: all;
+}
+
+.span > :global(button:disabled) {
```
martinkrulltott, line 40: what does `:global` do here?

src/pages/view/FilterBar/__tests__/FilterBadge.spec.js
```diff
@@ -24,15 +26,15 @@ test('Filter Badge displays badge containing number of items in filter', () => {
     expect(container).toMatchSnapshot()
 })
 
-test('FilterBadge displays badge with filter item name when only one filter item', () => {
+test.skip('FilterBadge displays badge with filter item name when only one filter item', () => {
```
martinkrulltott, line 29:
```suggestion
test('FilterBadge displays badge with filter item name when only one filter item', () => {
```

why skip?

src/pages/view/FilterBar/styles/FilterBadge.module.css
martinkrulltott: applies to a few other places as well. maybe just do a search and replace?

src/pages/view/DashboardsBar/styles/DashboardsBar.module.css
```diff
@@ -33,6 +37,16 @@
     height: var(--user-rows-height);
 }
 
+.offline {
+    background-color: rgb(195, 100, 100) !important;
```
martinkrulltott: why important? and where are these non-ui colors coming from?

src/pages/view/DashboardsBar/styles/Content.module.css
```diff
@@ -2,19 +2,24 @@
...
 }
 
 .controlsLarge {
+    display: inline-flex;
+    position: relative;
+    top: 5px;
+}
+
+.buttonPadding {
+    padding: 2px 8px 0 8px;
```
martinkrulltott:
```suggestion
    padding: 2px var(--spacers-dp8 0 var(--spacers-dp8;
```

src/pages/view/FilterBar/styles/FilterBadge.module.css
jenniferarnesen: I'm not sure, this was the css Hendrick gave me to resolve the issue of a disabled button that can have a tooltip.

src/pages/edit/ItemSelector/ItemSearchField.js
```diff
@@ -1,19 +1,39 @@
...
+    )
+
+    if (online) {
+        return <>{getInput()}</>
+    }
+
+    return (
+        <Tooltip
+            content={i18n.t('Cannot search for dashboard items while offline')}
+            openDelay={200}
```
martinkrulltott: I've seen these delays in multiple places, maybe they can be moved to be global setting consts somewhere?

src/pages/edit/FilterSettingsDialog.js
```diff
@@ -165,18 +166,24 @@ const FilterSettingsDialog = ({
...
-                                    if (!filtersSelectable) {
-                                        setSelected([])
-                                    }
-                                    onConfirm(filtersSelectable, selected)
-                                }}
-                                primary
-                                type="button"
+                            <Tooltip
+                                content={i18n.t(
+                                    'Cannot confirm changes while offline'
```
martinkrulltott, line 171: Shouldn't this tooltip conditionally rendered based on `!filtersSelectable` or the online status to not show when online?

src/pages/edit/ActionsBar.js
```diff
@@ -172,29 +172,54 @@ const EditBar = ({ dashboard, ...props }) => {
...
+                <Button
+                    disabled={!online}
+                    primary
+                    onClick={onSave}
+                    dataTest="save-dashboard-button"
+                >
+                    {i18n.t('Save changes')}
                 </Button>
+            </Tooltip>
+            <Tooltip>
```
martinkrulltott: why wrap in a tooltip if the tooltip doesn't have any content? same for the other ones below as well

martinkrulltott: ah, this is a custom component, not the ui tooltip. maybe it can be called OfflineTooltip instead?

src/components/__tests__/ConfirmActionDialog.spec.js
```diff
@@ -21,8 +18,10 @@ jest.mock('@dhis2/ui', () => {
 test('ConfirmActionDialog renders confirm delete dialog', () => {
     const { container } = render(
         <ConfirmActionDialog
-            action={ACTION_DELETE}
-            dashboardName="Twilight Sparkle"
+            title="Delete dashboard"
+            message="Deleting dashboard Twilight Sparkle will remove it for all users. This action cannot be undone. Are you sure you want to permanently delete this dashboard?"
```
martinkrulltott, line 22: how can this test pass when the implementation code has quotes around the name and the test doesn't?

> message={i18n.t(
                        'Deleting dashboard "{{ dashboardName }}" will remove it for all users. This action cannot be undone. Are you sure you want to permanently delete this dashboard?',
                        { dashboardName: dashboard.name }
                    )}

src/components/Item/VisualizationItem/ItemContextMenu/__tests__/ItemContextMenu.offline.spec.js
```diff
@@ -0,0 +1,479 @@
...
+    const { getByRole, queryByTestId, queryByText } = render(
+        <WindowDimensionsProvider>
+            <ItemContextMenu {...props} />
+        </WindowDimensionsProvider>
+    )
+
+    fireEvent.click(getByRole('button'))
+
+    await waitFor(() => {
+        expect(queryByText('View as Map')).toBeNull()
```
martinkrulltott, line 469: Since these are repeated across each step which makes the file pretty big, maybe they could be abstracted with a function like `expectViewAsMap = state => state ? expect(queryByText('View as Map')).toBeTruthy() : expect(queryByText('View as Map')).toBeNull()` to avoid repetition, or maybe even use a config for the tests, like `pieSmallScreen = { showViewAsMap: false, showViewAsChart: false, ... }`. To make it a bit easier to overview and maintain 🙏

martinkrulltott (commented):
Reviewed, see separate comments with suggestions and questions.

src/pages/view/TitleBar/LastUpdatedTag.js
jenniferarnesen: done

src/pages/view/TitleBar/LastUpdatedTag.js
jenniferarnesen: done

src/pages/view/FilterBar/styles/FilterBadge.module.css
jenniferarnesen: Fixed here. I'll do a repo wide fix in a separate PR.

src/pages/view/DashboardsBar/styles/DashboardsBar.module.css
jenniferarnesen: Removed. This was for when the offline/online state was being mocked.

src/pages/view/DashboardsBar/styles/Content.module.css
jenniferarnesen: done

src/pages/edit/FilterSettingsDialog.js
jenniferarnesen: The tooltip is for when the button is disabled due to being offline.

src/pages/edit/ActionsBar.js
jenniferarnesen: Yeah I was thinking that OfflineTooltip would be a better name.

src/components/__tests__/ConfirmActionDialog.spec.js
jenniferarnesen: The test isn't testing that i18n string. It is only a test of the ConfirmActionDialog, which just accepts a message property and displays it. (the code you quoted is from ActionsBar)

src/components/Item/VisualizationItem/ItemContextMenu/__tests__/ItemContextMenu.offline.spec.js
jenniferarnesen: I have usually done things the way you describe, but I read an article not too long ago about how to organize tests, and it recommended not pulling out common code into separate functions. Trade off is a lot of repetition, but on the other hand, it is much easier to understand the each test. That said, I'm not sure I'm in 100% agreement with the author. I'll try to find the article.

src/pages/edit/ActionsBar.js
jenniferarnesen: Done

src/pages/edit/FilterSettingsDialog.js
jenniferarnesen: I don't think so. I am only adding functionality for offline. Nothing else has changed unless I'm missing something?

edoardo (approved):

src/pages/edit/FilterSettingsDialog.js
martinkrulltott: hm, I thought, since this is wrapping the "confirm" button, which I thought would be visible both online and offline, that the tooltip would then also be (incorrectly) visible when online. The button has a `if (!filtersSelectable)` condition, but the tooltip has none. But maybe Im missing something, hard to judge when you just look at this in isolation.

PR #1727, opened 2021-04-27
fix: namespace check issue

janhenrikoverland:
Ref 0997714d89580c3b76a7e2510d760edbc3084155


jenniferarnesen:
The changes to the snapshots have to do with changes that happened to material-ui/icons some time ago (https://github.com/mui-org/material-ui/pull/16258). Not sure why the snapshot are only being updated now.

But as long as the icons in the Item context menu and Edit item selector look correct, then it's good.

jenniferarnesen (approved):

PR #1756, opened 2021-05-10
chore: changes to redux structure for dashboards and selected property

jenniferarnesen:
The 2 major changes in this PR are
1. change redux structure so that `selected` contains all the info (except `starred`) about the selected dashboard. `dashboards` prop in redux is now solely focused on the list of dashboards and their starred status.
2. Get rid of the `Dashboard` controller - the file that all routes (view, edit, new, print) went through when rendering the route. This was to decouple the routes from state that it doesn't depend on. (like Edit and New do not need the list of dashboards). The eventual purpose was to simplify the implementation of offline dashboard.

Previous redux structure:
```
dashboards: {
    byId: {all dashboards mapped by id},
    dashboardItems: {dashboardItems of selected dashboard}
},
selected: {
    id: [dashboardId],
    isLoading: false/true
}
```

New redux structure: 
```
dashboards: {all dashboards mapped by id},
selected: {
    id,
    displayName,
    displayDescription,
    dashboardItems,
    access,
    restrictFilters,
    allowedFilters
}
```

Notes:
* `getCustomDashboards` has been moved to the "source" - where the dashboard is fetched in api/fetchDashboard.
* `Item/Item` - just spread the props rather than listing each one separately. Makes it easier to add/change props later
* `EditDashboard`, `PrintDashboard`, `PrintLayoutDashboard` - all separately make the api request for fetch dashboard. I know this is a little bit of code duplication, but the number of requests hasn't changed (actually reduced by one since the edit ActionsBar no longer requests the dashboard)
* `PrintDashboard` and `PrintLayoutDashboard` are mostly the same except for the `loadDashboard` code in the effect. It just looks like a lot of changes, but mostly code has shifted down or up in the file.
* `CacheableViewDashboard`: This file is used to show loading status while dashboard list is loading, and to handle the route id for the `view` route. This separation of concerns (this file vs ViewDashboard) is needed for the upcoming offline dashboard.

There is already decent Cypress coverage of the changes in this PR (loading, switching dashboard, going in and out of edit/new/print modes). But I'll look for additional tests to add.


src/pages/print/PrintLayoutItemGrid.js
```diff
@@ -21,7 +17,7 @@ import { getDomGridItemsSortedByYPos, getTransformYPx } from './printUtils'
 import { PRINT_LAYOUT } from '../../modules/dashboardModes'
 import { PAGEBREAK } from '../../modules/itemTypes'
 
-export class PrintLayoutItemGrid extends Component {
+class PrintLayoutItemGrid extends Component {
```
edoardo, line 20: Convert to functional component?

edoardo (approved):
It seems fine, I tested a bit, but I guess this is a candidate for a testing session.

src/pages/print/PrintLayoutItemGrid.js
jenniferarnesen: Will take a look in a separate PR

PR #1769, opened 2021-05-18
fix: all cypress tests in edit mode should check that view mode is displayed first

jenniferarnesen:
Some tests occasionally fail because a button (e.g. Share) is not yet visible. This change adds a check for each test that ensures that view mode has been rendered before moving on with the rest of the test.

### DIFF
--- a/cypress/integration/ui/edit_dashboard/edit_dashboard.js
+++ b/cypress/integration/ui/edit_dashboard/edit_dashboard.js
@@ -95,6 +95,16 @@ Given('I open existing dashboard', () => {
     cy.get(dashboardChipSel, EXTENDED_TIMEOUT)
         .contains(TEST_DASHBOARD_TITLE)
         .click()
+
+    cy.location().should(loc => {
+        const currentRoute = getRouteFromHash(loc.hash)
+
+        expect(nonViewRoutes).not.to.include(currentRoute)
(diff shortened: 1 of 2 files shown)


cypress/integration/ui/edit_dashboard/edit_dashboard.js
```diff
@@ -95,6 +95,16 @@ Given('I open existing dashboard', () => {
     cy.get(dashboardChipSel, EXTENDED_TIMEOUT)
         .contains(TEST_DASHBOARD_TITLE)
         .click()
+
+    cy.location().should(loc => {
+        const currentRoute = getRouteFromHash(loc.hash)
+
+        expect(nonViewRoutes).not.to.include(currentRoute)
```
martinkrulltott, line 102: This works, but my spontaneous thought was to check that it's a valid uid (i.e. valid format, not valid in the db).

martinkrulltott: Actually, maybe that's what you're doing on the line below though? 🤔

martinkrulltott (approved):

cypress/integration/ui/edit_dashboard/edit_dashboard.js
jenniferarnesen: Yepp

PR #1789, opened 2021-05-28
feat: hide periods based on system settings (DHIS2-11161)

martinkrulltott:
Implements [DHIS2-11161](https://jira.dhis2.org/browse/DHIS2-11161)

**Requires https://github.com/dhis2/analytics/pull/920**

Relates to https://github.com/dhis2/data-visualizer-app/pull/1721

---

### Key features

1. Populates `excludedPeriodTypes` with values from system settings
2. Changes all instances of `DAYS` and `WEEKS` from Analytics to `DAILY` and `WEEKLY`

---

### Description

Fetches props from system settings to be passed to `excludedPeriodTypes`. Based on the DV change (https://github.com/dhis2/data-visualizer-app/pull/1721). 

---

### TODO

- [ ] Move `getExcludedPeriodTypes` to Analytics?

---

### Screenshots

_see the Analytics PR_

### DIFF
--- a/src/pages/view/TitleBar/FilterDialog.js
+++ b/src/pages/view/TitleBar/FilterDialog.js
@@ -17,6 +17,15 @@ import {
     OrgUnitDimension,
     DIMENSION_ID_PERIOD,
     DIMENSION_ID_ORGUNIT,
+    DAILY,
+    WEEKLY,
+    WEEKLYWED,
+    WEEKLYTHU,
+    WEEKLYSAT,
+    WEEKLYSUN,
+    BIWEEKLY,
+    MONTHLY,
+    BIMONTHLY,
 } from '@dhis2/analytics'
 import {
     acAddItemFilter,
@@ -24,6 +33,7 @@ import {
 } from '../../../actions/itemFilters'
 import { sGetItemFiltersRoot } from '../../../reducers/itemFilters'
 import { useUserSettings } from '../../../components/UserSettingsProvider'
+import { useSystemSettings } from '../../../components/SystemSettingsProvider'
 
 const FilterDialog = ({
     dimension,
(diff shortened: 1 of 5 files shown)


src/pages/view/TitleBar/FilterDialog.js
```diff
@@ -90,6 +121,9 @@ const FilterDialog = ({
                     <PeriodDimension
                         selectedPeriods={selectedItems}
                         onSelect={commonProps.onSelect}
+                        excludedPeriodTypes={getExcludedPeriodTypes(
```
jenniferarnesen, line 124: Is the idea with making excludedPeriodTypes configurable in PeriodDimension that the app may want to be even more restrictive than the system settings? Could the app also enforce a less restrictive list than the system settings?

martinkrulltott: In my initial version of this, the fetching of system settings and excluding of periods were all done in Analytics itself. But it was decided to move it out to `excludedPeriodTypes` and let the apps decide which periods to exclude, since we want to keep the lib flexible and unopinionated. Also, since the apps already fetch the system settings, we avoid having a separate backend call in Analytics which would also lead to a having a loading spinner when you open the period dimension. 
So by letting the apps fetch the settings (which the already do, these are just additional props to the existing api request), we avoid both the loading spinner, an additional api request and opinionating Analytics.

So to answer the question, yes apps could provide any restrictions they want, regardless of the system settings.

jenniferarnesen: ok, if that's approved by the product manager, then ok with me.

jenniferarnesen (approved):

src/pages/view/TitleBar/FilterDialog.js
martinkrulltott: If what is approved?

PR #1795, opened 2021-05-31
fix: avoid loading user/groups in interpretations DHIS2-10625 v34

edoardo:
Backport.
Solves the issue of potentially loading a huge list of users when fetching user groups.
See https://jira.dhis2.org/browse/DHIS2-10625


jenniferarnesen (approved):

jenniferarnesen:
@edoardo Even though the Jira issue is in the title, could you write a short summary in the description of what the backport is?

PR #1812, opened 2021-06-09
fix: dashboard crashes if an item is missing type [TECH-588]

jenniferarnesen:
One character fix in `itemTypes.js`

The rest is support for new cypress tests that mock a request with missing or unrecognized item types.

Look at `view_errors.feature` for a summary of what the new new cypress tests do. Some cypress changes are just moving functions to `common` so that all tests can access the function


src/components/Item/NotSupportedItem/Item.js
```diff
@@ -4,28 +4,35 @@ import i18n from '@dhis2/d2-i18n'
...
+              type: item.type,
+          })
+        : i18n.t('The item type is missing')
+
+    return (
+        <>
+            <ItemHeader
+                itemId={item.id}
+                dashboardMode={dashboardMode}
+                isShortened={false}
```
edoardo: Is `item.shortened` not available anymore? Or is it supposed to be always `false` for non supported items?

edoardo (approved):

src/components/Item/NotSupportedItem/Item.js
jenniferarnesen: Wait you are right. It could be shortened.

PR #1819, opened 2021-06-14
fix: include hideTitle property when loading map [DHIS2-11302]

jenniferarnesen:
Make sure to include `options` which contains `hideTitle` when loading the map

image


edoardo (approved):

jenniferarnesen:
Make sure to include "options" which contains "hideTitle" when loading the map

PR #1858, opened 2021-07-01
fix: expanded control bar not resizing on drag [DHIS2-10795] [v36]

jenniferarnesen:
Bug fix: remove the expanded flag if dragging has decreased the controlbar height to less than max height.

Backport of https://github.com/dhis2/dashboard-app/pull/1681

For the smoother resize experience:

* reduced re-rendering of DragHandle by using React.memo (this also results in a smoother drag experience since the initial startingY state doesn't get reset to 0)
* use Math.round instead of Math.floor

### DIFF
--- a/cypress/integration/ui/view_dashboard.feature
+++ b/cypress/integration/ui/view_dashboard.feature
@@ -54,10 +54,17 @@ Feature: Viewing dashboards
         When I toggle the control bar height
         Then the control bar should be expanded to full height
 
+# TODO: flaky test
+# @mutating
+# Scenario: I change the height of the control bar
+#     Given I open the "Delivery" dashboard
+#     When I drag to increase the height of the control bar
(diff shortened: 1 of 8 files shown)


cypress/integration/ui/view_dashboard.feature
```diff
@@ -54,10 +54,17 @@ Feature: Viewing dashboards
         When I toggle the control bar height
         Then the control bar should be expanded to full height
 
+# TODO: flaky test
```
edoardo, line 57: Is this something to fix?

edoardo (approved):

cypress/integration/ui/view_dashboard.feature
jenniferarnesen: yes, I need to take a look at all the flaky tests and figure out how to make them reliable.

PR #1864, opened 2021-07-06
fix: add onError callback to handle failed chart from DV plugin [DHIS2-11303] [v35]

jenniferarnesen:
Backport of https://github.com/dhis2/dashboard-app/pull/1826

Add callback for errors from DataVisualizerPlugin. The implementation is based on the sketch from DHIS2-2839.

image

In print mode:
image

PR #1865, opened 2021-07-06
fix: add onError cb to handle failed chart from DV plugin [DHIS2-11303] [v34]

jenniferarnesen:
Backport of #1826

Add callback for errors from DataVisualizerPlugin. The implementation is based on the sketch from DHIS2-2839.

image


edoardo (approved):
Same comment as in #1864

PR #1892, opened 2021-08-09
fix: bump dv plugin and ui to latest

martinkrulltott:
Bumps @dhis2/data-visualizer-plugin and @dhis2/ui to latest to implement the fix for [DHIS2-11239](https://jira.dhis2.org/browse/DHIS2-11239)


jenniferarnesen (approved):
ok

PR #1909, opened 2021-08-20
fix: replace icons used for list item delete items

jenniferarnesen:
Old
image

New
image


src/components/Item/ListItem/Item.js
```diff
@@ -11,8 +12,6 @@ import { isEditMode } from '../../../modules/dashboardModes'
 import { itemTypeMap, getItemUrl } from '../../../modules/itemTypes'
 import { orArray } from '../../../modules/util'
 import ItemHeader from '../ItemHeader/ItemHeader'
-import DeleteIcon from './assets/Delete'
-import DescriptionIcon from './assets/Description'
```
edoardo, line 15: Is this icon used somewhere else? I don't see it removed from the repo in this PR

edoardo (approved):

cooper-joe (approved):

src/components/Item/ListItem/Item.js
jenniferarnesen: Right. Removed.

PR #1912, opened 2021-08-24
fix: adjust z-index so Show more/fewer dashboards tooltip is visible [aDHIS2-11647]

jenniferarnesen:
https://jira.dhis2.org/browse/DHIS2-11647

Because of competing z-indexes on the component cover and tooltip, the tooltip was hidden when the dashboards bar was expanded.

Components should be stacked from bottom to top:  cover, dashboards bar, tooltip.
Previously, the z-index of these was: 2000, 2001, 2000 which means the tooltip wasn't visible.
With the new code, the z-index is: 1998, 1999, 2000. (Cannot use ui ComponentCover since z-index is not configurable).

Regular screen:

https://user-images.githubusercontent.com/6113918/130608062-db8da82b-5441-4208-9504-5d960c584228.mov

Small screen:

https://user-images.githubusercontent.com/6113918/130612293-5ceb9b59-c984-4469-8c77-9bd4f6875777.mov

### DIFF
--- a/src/pages/view/DashboardsBar/ShowMoreButton.js
+++ b/src/pages/view/DashboardsBar/ShowMoreButton.js
@@ -34,7 +34,12 @@ const ShowMoreButton = ({ onClick, dashboardBarIsExpanded, disabled }) => {
                     <ChevronDown />
                 </div>
             ) : (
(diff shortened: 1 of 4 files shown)


janhenrikoverland (approved):
LGTM

PR #1916, opened 2021-08-30
fix: migrate from chart/pt to visualizations

martinkrulltott:
updates needed to migrate from chart and pivot tables to visualizations, as the deprecated endpoints will be dropped for 2.37

Targeted towards the latest backend changes that can be tested here: https://prep.dhis2.org/DHIS2-11047/

or here:

https://test.e2e.dhis2.org/DHIS2-11047/

### DIFF
--- a/src/components/Item/VisualizationItem/Visualization/__tests__/Visualization.spec.js
+++ b/src/components/Item/VisualizationItem/Visualization/__tests__/Visualization.spec.js
@@ -57,8 +57,8 @@ test('renders a VisualizationPlugin for CHART', () => {
             <Visualization
                 item={{
                     id: 'rainbow',
-                    type: 'CHART',
-                    chart: { id: 'rainbowVis' },
+                    type: 'VISUALIZATION',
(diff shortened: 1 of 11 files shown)


jenniferarnesen (approved):

src/components/Item/VisualizationItem/Visualization/__tests__/Visualization.spec.js
```diff
@@ -75,8 +75,8 @@ test('renders a VisualizationPlugin for REPORT_TABLE', () => {
             <Visualization
                 item={{
                     id: 'rainbow',
-                    type: 'REPORT_TABLE',
-                    reportTable: { id: 'rainbowVis' },
+                    type: 'VISUALIZATION',
+                    visualization: { id: 'rainbowVis', type: 'PIVOT_TABLE' },
                 }}
                 activeType="REPORT_TABLE"
```
martinkrulltott, line 81: shouldn't these be changed to?

PR #1933, opened 2021-09-02
fix: require a number and only allow numbers greater than zero for layout columns (DHIS2-11703)

janhenrikoverland:
Currently the number of columns could be set to something lower than 1, which is nonsensical. Also, the number should not be erasable.


martinkrulltott (approved):
LGMT. Note that I could still go from fixed to freeflow by typing in 30 and removing the 3 (leaving 0, which switched it over to freeflow). But I think we can live with that 🙂 
A bit unusual to not be able to remove the value though, are we sure all browsers provide up and down arrows to change the value? (otherwise to go from 3 to 4 you'd have to e.g. type 34 and then remove the 3).

PR #1940, opened 2021-09-07
fix: add callback for when recording fails

jenniferarnesen:
The callback does not trigger in cases of 40x from the backend.

Update:
image

https://user-images.githubusercontent.com/6113918/132322509-51ff5e06-16b1-4d3a-855a-c09c9dbab62e.mov

### DIFF
--- a/src/pages/view/TitleBar/ActionsBar.js
+++ b/src/pages/view/TitleBar/ActionsBar.js
@@ -70,29 +70,35 @@ const ViewActions = ({
         return <Redirect to={redirectUrl} />
     }
 
+    const onRecordError = () => {
+        warningAlert.show({
+            msg: i18n.t('The dashboard could not be cached. Please try again'),
+        })
+    }
+
     const onCacheDashboardConfirmed = () => {
         setConfirmCacheDialogIsOpen(false)
         removeAllFilters()
-        startRecording({})
+        startRecording({
+            onError: onRecordError,
+        })
     }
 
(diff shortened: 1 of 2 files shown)


edoardo (approved):

src/pages/view/TitleBar/ActionsBar.js
```diff
@@ -70,29 +70,35 @@ const ViewActions = ({
         return <Redirect to={redirectUrl} />
     }
 
+    const onRecordError = () => {
+        warningAlert.show({
+            msg: i18n.t('The dashboard could not be cached. Please try again.'),
```
cooper-joe: For consistency, I think the message `The dashboard couldn't be made available offline. Try again.` is clearer, as _caching_ is not referred to anywhere else.

cooper-joe (changes requested):

src/pages/view/TitleBar/ActionsBar.js
jenniferarnesen: @cooper-joe Should it also be a critical message rather than a warning?

jenniferarnesen: I updated the PR description with a screenshot showing the new text plus using critical

cooper-joe: Looks good 👍

cooper-joe (approved):

PR #1945, opened 2021-09-07
fix: show/hide description regardless of whether offline

jenniferarnesen:
fixes https://jira.dhis2.org/browse/DHIS2-11723

Note that once the user shows/hides while offline, then the database and app may be out of sync until the next time the user reloads the app or changes the setting once online again.

Previously there was a warning message if the api request for showDescription preference failed, but that has been removed, per UX.

In this video, the user is offline:

https://user-images.githubusercontent.com/6113918/132518752-5640ac73-ed85-4397-9bf7-60a064ce5bf4.mov

### DIFF
--- a/cypress/integration/view/offline.feature
+++ b/cypress/integration/view/offline.feature
@@ -65,6 +65,12 @@ Feature: Offline dashboard
         When connectivity is turned off
         Then it is not possible to interact with interpretations
 
+    Scenario: I show the description while offline
+        Given I open an uncached dashboard
+        And connectivity is turned off
+        When I choose Show Description
+        Then the description is shown along with a warning
+
 
     Scenario: I delete the cached and uncached dashboard
         Given I delete the cached and uncached dashboard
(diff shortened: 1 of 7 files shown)


cooper-joe:
I don't think a warning should be shown. This is a very non-critical action that doesn't have any serious consequences, just that the app isn't going to remember the show/hide state of the description. I think showing a warning here would:
- potentially scare a user into thinking they've done something wrong.
- desensitize users to think that warnings are shown for non-consequential issues.

So I'm in favour of simply not showing the warning.

jenniferarnesen:
> I don't think a warning should be shown. This is a very non-critical action that doesn't have any serious consequences, just that the app isn't going to remember the show/hide state of the description. I think showing a warning here would:
> 
> * potentially scare a user into thinking they've done something wrong.
> * desensitize users to think that warnings are shown for non-consequential issues.
> 
> So I'm in favour of simply not showing the warning.

Agreed

edoardo (approved):

PR #1985, opened 2021-09-16
fix: restore dashboard filter tests [v37]

jenniferarnesen:
These tests had been temporarily removed due to a bug in dependency. So the files `dashboard_filter.js` and `dashboard_filter.feature` are pretty much unchanged.

Some extended timeouts were added to a few tests that were acting flaky, and a test dashboard name was shortened.

### DIFF
--- a/cypress/integration/view/offline/offline.js
+++ b/cypress/integration/view/offline/offline.js
@@ -175,10 +175,12 @@ Given('I open an uncached dashboard', () => {
 
 When('connectivity is turned off', () => {
     goOffline()
+    cy.wait(500) // eslint-disable-line cypress/no-unnecessary-waiting
 })
 
 When('connectivity is turned on', () => {
     goOnline()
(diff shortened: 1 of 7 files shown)


cypress/integration/view/offline/offline.js
```diff
@@ -175,10 +175,12 @@ Given('I open an uncached dashboard', () => {
 
 When('connectivity is turned off', () => {
     goOffline()
+    cy.wait(500) // eslint-disable-line cypress/no-unnecessary-waiting
```
martinkrulltott: Any way to get around the `wait`?

martinkrulltott (approved):

cypress/integration/view/offline/offline.js
jenniferarnesen: Could potentially wait until one of the action buttons becomes enabled. I can try that out.

PR #1987, opened 2021-09-16
fix: bump Analytics and DV plugin to latest (DHIS2-11017) [v36]

martinkrulltott:
Deps bump for [DHIS2-11017](https://jira.dhis2.org/browse/DHIS2-11017), to run [Analytics 16.0.22](https://github.com/dhis2/analytics/pull/1043) (which contains the actual fix) and [DV plugin 36.1.6](https://github.com/dhis2/data-visualizer-app/pull/1885) that uses the new Analytics version as well.

### DIFF
--- a/i18n/en.pot
+++ b/i18n/en.pot
@@ -5,8 +5,8 @@ msgstr ""
 "Content-Type: text/plain; charset=utf-8\n"
 "Content-Transfer-Encoding: 8bit\n"
 "Plural-Forms: nplurals=2; plural=(n != 1)\n"
-"POT-Creation-Date: 2021-06-10T13:31:39.680Z\n"
-"PO-Revision-Date: 2021-06-10T13:31:39.680Z\n"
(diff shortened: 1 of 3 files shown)


i18n/en.pot
```diff
@@ -92,14 +92,8 @@ msgstr ""
 msgid "Print"
 msgstr ""
 
-msgid "There was an error loading data for this item"
-msgstr "There was an error loading data for this item"
-
-msgid "Open this item in {{appName}}"
-msgstr "Open this item in {{appName}}"
-
-msgid "Visualizations"
```
jenniferarnesen, line 101: Do you know the source of these translation changes?

jenniferarnesen (approved):

i18n/en.pot
martinkrulltott: No, but the last revision date is 2021-06-10 so I assume its another commit between then and now. From the commit history it looks like most of them are actually yours, so maybe you recognise the strings?

PR #1997, opened 2021-09-27
fix: disable View as Chart/Table for non-thematic maps

jenniferarnesen:
Fixes https://jira.dhis2.org/browse/DHIS2-11889

Fixes in this PR
- Open in [appName] app should be enabled even when offline
- When online, View as options for non-thematic maps should be disabled with tooltip "This map not available as a Chart"
- Dimension filter settings confirm button should not have an offline tooltip, because clicking the "Confirm" button doesn't trigger an api request

Online, non-thematic map on the left:

https://user-images.githubusercontent.com/6113918/135046817-517ea37c-576f-4829-98ab-56f0f34600f1.mov

Offline - Open in Maps app should be enabled

https://user-images.githubusercontent.com/6113918/135046864-75c095cc-8dc7-4d63-bb13-a5871debb6f8.mov

Offline, non-thematic map - message should be "Not available offline"

https://user-images.githubusercontent.com/6113918/135046920-ba97aac3-f701-4eb3-bb94-4734045e825b.mov

### DIFF
--- a/cypress/integration/view/offline/offline.js
+++ b/cypress/integration/view/offline/offline.js
@@ -211,7 +211,7 @@ Then(
 
         cy.contains('li', 'View as').should('have.class', 'disabled')
         cy.contains('li', 'Open in Data Visualizer app').should(
-            'have.class',
+            'not.have.class',
             'disabled'
         )
         cy.contains('li', 'Show details and interpretations').should(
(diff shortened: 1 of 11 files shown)


cooper-joe (approved):
Design/texts LGTM 👍

edoardo (approved):

PR #2009, opened 2021-10-04
fix: store data that is specific to individual dashboard in dashboard-specific cache instead of app cache

jenniferarnesen:
The list of endpoints to NOT store in the app cache should include requests to fetch a map, ev, er, map tiles, and messages. They should instead be should stored in the specific dashboard cache.

The user won't notice this change. Either way, these requests are still cached, but this is the cleaner, and intended, solution.


d2.config.js
```diff
@@ -10,9 +10,14 @@ const config = {
             patternsToOmit: [
                 'dashboards/[a-zA-Z0-9]*',
                 'visualizations',
+                'maps',
```
KaiVandivier, line 13: Does this exclude the lazy-loaded `maps.js` from the app cache? It might need to be in there

KaiVandivier: Yeah I think this would match /dhis-web-**maps**/map.js, which we want in the app-shell cache because it's not guaranteed to be fetched and recorded when the dashboard is reloaded for recording. A more specific regex is necessary -- maybe `/\/maps\//`? (would match `/maps/etc` urls)

PR #2041, opened 2021-10-18
fix: if map contains an ee layer, then do not display it when offline [v37]

jenniferarnesen:
Updated fix for https://jira.dhis2.org/browse/DHIS2-11908

Instead of showing partial maps (e.g. maps with all its layers except ee layers) when offline, do not show the map at all if it has an EE layer. There was a lot of complexity and potential bugs with the previous solution, which was removing ee layers and displaying the rest of the layers.

https://user-images.githubusercontent.com/6113918/137713141-3faf8ac9-4957-425c-a9a7-ad80e5814e7d.mov


janhenrikoverland (approved):
LG

PR #2053, opened 2021-11-04
fix: make landing page scrollable

janhenrikoverland:
Makes the landing page scrollable. To enable a scrollbar the container needs a height, which in turn needs to be dynamic, so the height of the container is set by using flex-grow = 1 to make it fill the rest of the viewport height. To be able to use flex-grow the container's container needs to be flex. The new overflow rule is set on the StartScreen component so that it does not touch other stuff like dashboard print view.

Web:
https://user-images.githubusercontent.com/1010094/140347833-56e46d7d-51ce-41b3-9327-55df48611926.mp4

Mobile:
https://user-images.githubusercontent.com/1010094/140348329-a44247cf-c2c7-4273-b23b-f699549dc769.mp4

### DIFF
--- a/src/pages/start/styles/StartScreen.module.css
+++ b/src/pages/start/styles/StartScreen.module.css
@@ -1,7 +1,8 @@
 .outer {
     display: flex;
     justify-content: center;
-    height: 85%;
+    flex-grow: 1;
+    overflow: auto;
 }
 .inner {
     padding: var(--spacers-dp16) var(--spacers-dp24);
@@ -9,8 +10,7 @@
     box-sizing: border-box;
     color: var(--colors-grey900);
     position: relative;
-    display: flex;
-    align-items: center;
+    padding-top: 15vh;
 }
 .errorContainer {
     display: flex;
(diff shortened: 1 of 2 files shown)


src/pages/start/styles/StartScreen.module.css
```diff
@@ -1,16 +1,16 @@
...
 }
 .inner {
     padding: var(--spacers-dp16) var(--spacers-dp24);
     max-width: 600px;
     box-sizing: border-box;
     color: var(--colors-grey900);
     position: relative;
-    display: flex;
-    align-items: center;
+    padding-top: 15vh;
```
martinkrulltott, line 13: where is this coming from? is this the diff from the "85%" that was removed?

martinkrulltott (approved):

src/pages/start/styles/StartScreen.module.css
janhenrikoverland: I can't get scrolling to work the way we want it if we have any kind of vertical centering. A dynamic buffer distance works okay.

PR #2057, opened 2021-12-20
feat: dashboard code (DHIS2-11790)

janhenrikoverland:
Adds support for saving a code with the dashboard. Code is already part of the dashboard entity (identifiable object) so no changes needed in the backend.

Code must be unique, so a message is shown to the user and saving of the dashboard is blocked if the provided code is already used by another dashboard. To separate this use case from a regular "save failed" we're looking for the internal error code E5003 when a 409 is returned by the server.

The case where a dashboard is updated with the same code (which should be allowed) is handled by the backend (PUT request looking at the id), so that works out of the box.

Code can't be longer than 50 characters.

Feedback wording by @cooper-joe 

Screenshot from 2021-12-20 11-45-50

Screenshot from 2021-12-20 12-10-39


src/pages/edit/styles/TitleBar.module.css
```diff
@@ -18,6 +18,16 @@
...
 }
 
+.inputFieldWrapper {
+    display: flex;
+    flex-direction: column;
+    justify-content: space-between;
+}
+
+.inputFieldWrapper > :not(:first-child) {
+    padding-top: 8px;
```
jenniferarnesen: var(--spacers-dp8)

src/pages/edit/ActionsBar.js
```diff
@@ -33,6 +33,10 @@ const saveFailedMessage = i18n.t(
     'Failed to save dashboard. You might be offline or not have access to edit this dashboard.'
 )
 
+const saveFailedCodeExistsMessage = i18n.t(
+    'This code is already being used on another dashboard'
```
jenniferarnesen, line 37: For consistency:  "Failed to save dashboard. Code is already being used on another dashboard."

src/pages/edit/styles/TitleBar.module.css
janhenrikoverland: Thanks

src/pages/edit/ActionsBar.js
janhenrikoverland: @cooper-joe ^

jenniferarnesen (approved):

src/pages/edit/ActionsBar.js
cooper-joe: I don't think "failed to save dashboard" works here, because it's used as an error message underneath a field. The other error (about being offline) is used in an alert bar.

janhenrikoverland: @cooper-joe No, this is the alert bar message after trying to save an already used code, I think you mix it up with "Code can't be longer than 50 characters" which is the client side check on the input field. Unless the intention was to display both of them as an input field error? See screenshots in the PR.

cooper-joe: Oh yeah, sorry about that. Agreed with @jenniferarnesen! 

I do also think the _unique code_ error should also display as an input field error, as well as an alert bar communicating the failed save.

janhenrikoverland: Okay, interesting. That could work, but there are some questions related to that, e.g. for how long should that message remain visible? Same as the alert bar, until the field is edited, semi-permanently until the user leaves the new/edit screen? If the user edits the code the message is no longer correct, unless the user understands that this is only updated after going to the server (not something users should have to consider, imo), as opposed to the max 50 chars check which is instant. As @cooper-joe is on holiday as we speak I'm going to merge the PR for now and revisit this in January.

janhenrikoverland: Forgot to update the message, will create a new PR for that.

PR #2059, opened 2022-01-10
fix: adjust z-index so tooltip shows [DHIS2-11647] [v36]

jenniferarnesen:
Backport of https://github.com/dhis2/dashboard-app/pull/1912

Fixes: https://jira.dhis2.org/browse/DHIS2-11647

Because of competing z-indexes on the component cover and tooltip, the tooltip was hidden when the dashboards bar was expanded.

Components should be stacked from bottom to top: cover, dashboards bar, tooltip.
Previously, the z-index of these was: 2000, 2001, 2000 which means the tooltip wasn't visible.
With the new code, the z-index is: 1998, 1999, 2000. (Cannot use ui ComponentCover since z-index is not configurable).


janhenrikoverland (approved):
LG

PR #2074, opened 2022-03-02
chore(release): migrate to new app version scheme

varl:
Align all our application versions that range between 0.x.x and 36.x.x
with different semantics to a single version scheme with distinct
semantics for what constitutes a breaking change.

We have chosen to bump to v100.0.0 to signify the depature from the old
version scheme to the new.

For more information:
dhis2/notes#293

BREAKING CHANGE: App version becomes decoupled from DHIS2 versions, see
the d2.config.js or App Hub for DHIS2 version compatibility.

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -1,0 +1,6 @@
+# default code owner for repo
+*        @dhis2/team-analytics
+
+# code owners for functionality
+/src/    @dhis2/team-analytics @dhis2/team-qa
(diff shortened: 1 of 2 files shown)


varl:
@amcgee could you follow this up ? :radio:

amcgee:
Yeah thanks for the ping @varl!  @janhenrikoverland should we sync on this some time this week?

PR #2075, opened 2022-03-15
fix: do not include fields more than once in api request

jenniferarnesen:
Fixes: https://jira.dhis2.org/browse/DHIS2-12859

A recent change in the api revealed that dashboard was including the `favorite` field twice in the GET request for dashboard when editing. Solution is to make sure to only include the field once

https://user-images.githubusercontent.com/6113918/158381666-e49a7e11-145c-481d-8f73-3a5b713304ed.mov

.


turban (approved):
LGTM!

PR #2110, opened 2022-09-19
refactor: iframe plugin support

edoardo:
**Fixes [DHIS2-14216](https://dhis2.atlassian.net/browse/DHIS2-14216)**

### Key features

1. Extend the dashboard item search to include the `event visualizations`.

2. Implement the new secure iframe plugin solution needed for LL, DV and Map plugins and the other ones that will follow.
These dashboard items are rendered as `iframe` tags pointing to the location of the plugin.
Normally the URL used is the one specified in `pluginLaunchUrl` returned by the `apps` API.
There is a backdoor which is enabled only in dev mode where it's possible to override those locations via `localStorage` for development/testing purposes (ie. pointing a plugin to a port on localhost where the plugin runs).

Here's an example of override for both DV and LL plugins:
Screenshot 2022-11-09 at 16 17 47

[DHIS2-14216]: https://dhis2.atlassian.net/browse/DHIS2-14216?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/components/Item/VisualizationItem/ItemContextMenu/ViewAsMenuItems.js
```diff
@@ -47,19 +48,23 @@ const ViewAsMenuItems = ({
...
-                            ? i18n.t("This map can't be displayed as a table")
-                            : null
-                    }
-                    label={i18n.t('View as Table')}
-                    onClick={onViewTable}
-                    disabled={notSupported}
-                    icon={<IconTable16 />}
-                />
-            )}
+            {activeType !== REPORT_TABLE &&
```
jenniferarnesen: With the 3rd check, maybe better to switch to 
`![REPORT_TABLE, EVENT_REPORT, EVENT_VISUALIZATIOn].includes(activeType)`

What do you think?

src/components/Item/VisualizationItem/Visualization/Visualization.js
```diff
@@ -21,100 +22,113 @@ import memoizeOne from '../memoizeOne.js'
...
+    const memoizedGetFilteredVisualization = memoizeOne(
+        getFilteredVisualization
+    )
+    const memoizedGetVisualizationConfig = memoizeOne(getVisualizationConfig)
 
-        this.memoizedGetFilteredVisualization = memoizeOne(
-            getFilteredVisualization
-        )
-        this.memoizedGetVisualizationConfig = memoizeOne(getVisualizationConfig)
+    const getFilterVersion = memoizeOne(() => uniqueId())
```
jenniferarnesen: Can you confirm that the memoization is still working correctly? I seem to remember leaving this component as a class because of issues with memoization. But I could be mis-remembering.

jenniferarnesen: Also, if switching to functional component, I wonder if `useMemo` would work.

edoardo: I didn't test the filters yet, since we don't know if we need to disable them for LL.
But good pointing this out, I wouldn't have known.

src/components/Item/VisualizationItem/Visualization/IframePlugin.js
```diff
@@ -0,0 +1,106 @@
...
+    const iframeRef = useRef()
+    const [error, setError] = useState(false)
+
+    const onError = () => setError(true)
+
+    useEffect(() => {
+        const listener = postRobot.on(
+            'getProps',
+            // listen for messages coming only from the iframe rendered by this component
+            { window: iframeRef.current.contentWindow },
```
jenniferarnesen: Any reason to be concerned that `iframRef` wouldn't exist yet?

src/components/Item/VisualizationItem/Visualization/IframePlugin.js
```diff
@@ -0,0 +1,106 @@
...
+                <VisualizationErrorMessage
+                    item={item}
+                    dashboardMode={dashboardMode}
+                />
+            </div>
+        )
+    }
+
+    return (
+        <Suspense fallback={<div />}>
```
jenniferarnesen: Will there be a loading spinner like the other visualizations?

src/components/Item/VisualizationItem/Visualization/Visualization.js
jenniferarnesen: Actually, I think one of the issues could have been the amount of rendering. If you edit a dashboard and resize or move an item, is there unnecessary re-rendering/painting of any of the visualizations going on?

jenniferarnesen: Also look for unnecessary re-rendering in view mode when changing View As, opening the interpretations, etc.

src/components/Item/VisualizationItem/Visualization/IframePlugin.js
edoardo: I think the plugin itself might already have a spinner.
But I don't think we need `Suspense` here, the iframe tag can be rendered right away, the problem is that until the plugin is mounted and got all the props it needs, nothing will be rendered inside the iframe.
This might not be a problem as it does not even depend on network, but there might be other reasons why the iframe content takes time before rendering anything.

src/components/Item/VisualizationItem/Visualization/IframePlugin.js
edoardo: Once `Suspense` is removed I don't think so.

src/components/Item/VisualizationItem/ItemContextMenu/ViewAsMenuItems.js
edoardo: Changed.

src/components/Item/VisualizationItem/Visualization/IframePlugin.js
edoardo: Removed `Suspense`

src/components/Item/VisualizationItem/Visualization/IframePlugin.js
edoardo: Removed `Suspense`

edoardo: Clarification:
the new `PluginWrapper` components in the various apps show a loading spinner while the communication between dashboard app and iframe plugin content happen.

src/pages/edit/NewDashboard.js
```diff
@@ -2,9 +2,10 @@ import i18n from '@dhis2/d2-i18n'
 import cx from 'classnames'
 import PropTypes from 'prop-types'
 import React, { useState, useEffect } from 'react'
-import { connect } from 'react-redux'
+import { connect, useDispatch } from 'react-redux'
```
jenniferarnesen, line 5: What about switching connect with useSelector, for consistency?

PR #2111, opened 2022-09-22
fix: use new TranslationDialog component

edoardo:
**Requires https://github.com/dhis2/analytics/pull/1342**

Replace the old `d2-ui` component with the new one from analytics.

### Screenshot
Screenshot 2022-09-22 at 13 58 15


src/pages/edit/ActionsBar.js
```diff
@@ -159,23 +157,13 @@ const EditBar = ({ dashboard, ...props }) => {
...
-                }}
+                onClose={toggleTranslationDialog}
+                objectToTranslate={dashboard}
                 fieldsToTranslate={['name', 'description']}
-                onTranslationError={(err) =>
-                    console.log('translation update error', err)
-                }
                 onTranslationSaved={Function.prototype}
-                insertTheme={true}
-                isOnline={online}
```
jenniferarnesen, line 178: Does the new dialog support the offline disabling of the Save/Cancel buttons?

jenniferarnesen:
Could you add a screenshot to the description?

jenniferarnesen:
Remember to yarn remove package d2-ui-translation-dialog

src/pages/edit/ActionsBar.js
edoardo: Good point.
I don't think the current implementation of `TranslationDialog` has offline support, but it won't be difficult to add.
I think the `useOnlineStatus` hook can be used directly in the component in `analytics`.

edoardo: Added offline support for `TranslationDialog` in `analytics`.
Because the `OfflineTooltip` component is needed there, it's been moved from the app to `analytics` and used from there.

jenniferarnesen:
I found a bug. To reproduce:
1. Go into dashboard edit mode and click "Translate" (new translation dialog opens).
2. In dev tools Network tab, turn off network ("No throttling" -> "Offline").
3. An error appears in the alert bar at the bottom

image

image

jenniferarnesen (changes requested):
See screenshots of error

jenniferarnesen (approved):

PR #2117, opened 2022-09-27
chore: github workflow file and npm scripts updates based on convention

jenniferarnesen:
With this update dashboard now has the same verify configuration as line-listing-app.

Workflow files:
Line-listing-app workflow files: https://github.com/dhis2/line-listing-app/tree/dev/.github/workflows
Dashboard-app workflow files: https://github.com/dhis2/dashboard-app/tree/chore/update-gh-workflow-files/.github/workflows
(dashboard has an additional workflow for commenting and closing gh issues that may get opened)

The failing tests are due to the pivot table bug.


jenniferarnesen:
Will revisit this topic after a full workflow review by the platform team

PR #2144, opened 2022-10-30
fix: disable network checks [do not merge]

KaiVandivier:
This is a hotfix for local deployments that run into errors while disconnected from the internet (https://dhis2.atlassian.net/browse/DHIS2-12937)

The strategy is to make a fake `useOnlineStatus` hook that always returns `true` (and a fake `<OfflineTooltip>` component because that consumes `onlineStatus` too), then import the fakes in the existing files -- see `/src/online-status-fakes/`

This will be improved in the future by https://dhis2.atlassian.net/browse/LIBS-315

The idea is to make a bundle of this branch and include it in a custom WAR file for those offline deployments


KaiVandivier:
Closed in favor of #2145

PR #2145, opened 2022-10-31
fix: disable-network-checks using patch-package

jenniferarnesen:
This PR is an alternative to https://github.com/dhis2/dashboard-app/pull/2144. It shouldn't be necessary to do anything with the OfflineTooltip, since that component also uses the useOnlineStatus hook. This will also ensure that any other dependencies (i.e. from d2-ui, analytics, etc) that depend on online status will also get online: true.


KaiVandivier:
Looks great! This is a textual approval to avoid this getting merged accidentally 😁 ✅

I tested this pretty extensively both in a dev environment and a production build deployed on a local instance with the internet entirely disconnected.

jenniferarnesen:
This has been merged separately into a custom patch. Keeping the branch incase it is needed in the future.  
https://dhis2.atlassian.net/browse/LIBS-315

PR #2146, opened 2022-11-08
fix: bump ui dep with fix for cascade sharing (DHIS2-14361)

edoardo:
Actual bug fixed in https://github.com/dhis2/ui/pull/1177.

Before:
Screenshot 2022-11-08 at 15 24 38

After:
Screenshot 2022-11-08 at 13 54 58


KaiVandivier (approved):
Looks good :)

PR #2189, opened 2022-12-22
refactor: use native MessageChannel in place of postRobot

edoardo:
Experimental...


edoardo:
The plugin wrapper and sender components are going to be implemented in app-runtime and we are going to stick to postRobot, closing this PR.

PR #2199, opened 2023-01-05
fix: attempt at fixing re-render when interpretations panel toggles

edoardo:
The whole thing is quite complex, but basically the culprit is the `item` prop which changes its `h`(height) attribute when the interpretations panel is toggled.
This updated `item` object is required by React grid.
However, anything else in that object does not change when toggling the interpretations panel and re-render of the "visualizations" loaded in the item components is not necessary and not desired as it causes loading spinner and/or flashing of the visualization content.

The fix is basically to avoid relying on the whole `item` object in components that should not re-render when the item's height change.

Also, the `IframePlugin` is now wrapped in a `React.memo` and that required memoizing all the values for the props passed to the plugin.

Screenshots to show that the Visualization/IframePlugin components do not reload when the interpretations panel is toggled:

Screenshot 2023-01-05 at 15 07 18
Screenshot 2023-01-05 at 15 07 25

### DIFF
--- a/src/components/Item/VisualizationItem/Visualization/IframePlugin.js
+++ b/src/components/Item/VisualizationItem/Visualization/IframePlugin.js
@@ -18,12 +18,12 @@ import VisualizationErrorMessage from './VisualizationErrorMessage.js'
 const IframePlugin = ({
     activeType,
     filterVersion,
-    //    item,
     style,
     visualization,
     dashboardMode,
     dashboardId,
     itemId,
+    itemType,
 }) => {
     const { d2 } = useD2()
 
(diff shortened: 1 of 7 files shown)


jenniferarnesen (approved):

src/components/Item/VisualizationItem/Visualization/IframePlugin.js
```diff
@@ -146,14 +146,15 @@ const IframePlugin = ({
...
+                    itemType={itemType}
                     dashboardMode={dashboardMode}
                 />
             </div>
         ) : (
             <div style={style}>
                 <VisualizationErrorMessage
-                    //item={item}
+                    itemType={itemType}
+                    visualizationId={visualization.id}
```
jenniferarnesen, line 157: Is this the same as what was being calculated with `getVisualizationId` before?

jenniferarnesen (approved):

src/components/Item/VisualizationItem/Visualization/IframePlugin.js
edoardo: Yes:
https://github.com/dhis2/dashboard-app/pull/2199/files#diff-009a33196703f2c1b54a4360314f7e91351dc9b2323826f8a776c13d98ddc7a5L30
and
https://github.com/dhis2/dashboard-app/blob/master/src/modules/item.js#L21

Which is the AO in the `item` object, so the `id` is the AO's `id'.

PR #2228, opened 2023-03-01
fix: map plugin configuration

jenniferarnesen:
Maps need different handling when filtering since filters only apply to some map layers (thematic and event).


edoardo (commented):
Doesn't the module.css rename break `DataVisualizerPlugin.js` [here](https://github.com/dhis2/dashboard-app/blob/feat/map-plugin-latest/src/components/Item/VisualizationItem/Visualization/DataVisualizerPlugin.js#L5)?

jenniferarnesen:
> Doesn't the module.css rename break `DataVisualizerPlugin.js` [here](https://github.com/dhis2/dashboard-app/blob/feat/map-plugin-latest/src/components/Item/VisualizationItem/Visualization/DataVisualizerPlugin.js#L5)?

Is DataVisualizerPlugin even in use anymore?

PR #2254, opened 2023-03-19
fix: consolidate legacy and default plugin and fix resizing

jenniferarnesen:
This PR fixes 2 things:
- the offline message was not showing on EE maps. A blank item was shown, leading user to think the map just wasn't loading.
- The legacy plugin (old ER and EV) was not resizing in edit mode.
- Refactor: no longer need DefaultPlugin (merged together with LegacyPlugin) and MapPlugin (everything now handled in Visualization.js).

image


edoardo (approved):
It looks i18n.pot is loosing some strings that are needed (no filters, install app, plugin not available).

There's 1 lint error to fix.

PR #2285, opened 2023-04-25
feat(iframe-plugin): receive pwa installation status from plugins

KaiVandivier:
Receives and plugin installation status and uses it to load first one plugin of each time to allow its service worker to install, then load the rest of the plugins of that type from the cache to save network resources

Implements [DHIS2-15097](https://dhis2.atlassian.net/browse/DHIS2-15097)

Pairs with https://github.com/dhis2/data-visualizer-app/pull/2273

---

### TODO

- [x] At the app level: make a store of plugin types and their installation states. Each state is initialized to `'unknown'`
    - [x] As plugins send their installation states, update this store
- [x] At the dashboard level, get the _first_ plugin of each type within that dashboard
- [x] In each `IframePlugin` component, check the installation status for this plugin type:
    - [x] If the status == `'unknown'` or `'installing'`, **only** load the iframe if this is the first plugin of its type on this dashboard. Otherwise, wait with a loading UI
    - [x] If the status == `'ready'` or `null` (i.e., "won't load"), go ahead and load the iframe

---

### Known issues

-   May cause incompatibility with some plugin versions

---

### Screenshots

Before and after these changes plus the other plugin changes - on a local 2.40dev instance, with "Disable cache" in dev tools, on the ANC dashboard, and after using "Clear site data" in the Chrome/Brave dev tools:

anc-v40-original-load-perf

anc-v40-all-changes-load-perf

[DHIS2-15097]: https://dhis2.atlassian.net/browse/DHIS2-15097?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/Item/VisualizationItem/Visualization/IframePlugin.js
+++ b/src/components/Item/VisualizationItem/Visualization/IframePlugin.js
@@ -26,6 +26,9 @@ const IframePlugin = ({
     itemId,
     itemType,
 }) => {
+    // TODO: Access plugin installation status
+    // TODO: Access 'first plugins of their type' for this dashboard
+
     const { d2 } = useD2()
     const { baseUrl } = useConfig()
 
@@ -151,6 +154,8 @@ const IframePlugin = ({
                     window: iframeRef.current.contentWindow,
                 },
                 (event) => {
+                    // TODO: Update plugin installation status for this type
+
                     console.log(
                         'got installationStatus message; data:',
                         event.data
@@ -184,6 +189,12 @@ const IframePlugin = ({
             </div>
         )
     }
+
+    // TODO: if installationStatus === 'UNKNOWN' or 'INSTALLING':
+    // render iframe if this is the first plugin of its type on this dashboard.
+    // otherwise, render a loading spinner
+
+    // TODO: if installationStatus === 'READY' or null, load iframe normally
 
     return (
         <div className={classes.wrapper}>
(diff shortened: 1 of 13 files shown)


src/pages/print/PrintLayoutItemGrid.js
```diff
@@ -57,8 +58,16 @@ class PrintLayoutItemGrid extends Component {
         )
     }
 
-    getItemComponents = (items) =>
-        items.map((item) => this.getItemComponent(item))
+    getItemComponents = (items) => {
+        const firstOfTypes = getFirstOfTypes(items)
+        items.forEach((item) => {
```
edoardo: Perhaps this can be added to the map below to save one loop through all items...

edoardo (approved):

src/components/Item/VisualizationItem/Visualization/IframePlugin.js
```diff
@@ -1,15 +1,24 @@
 import { useConfig } from '@dhis2/app-runtime'
 import { useD2 } from '@dhis2/app-runtime-adapter-d2'
+import { Layer, CenteredContent, CircularLoader } from '@dhis2/ui'
```
KaiVandivier:
```suggestion
import { CenteredContent, CircularLoader } from '@dhis2/ui'
```

This is giving me an eslint warning locally

PR #2293, opened 2023-05-02
fix: disable LL items if the plugin version does not support installationStatus message

jenniferarnesen:
This is a workaround until a more robust system of plugin version management has been decided on. Without this fix, only the first LL item will load. The rest get an infinite spinner.

With this check:

image (3)

Without the check:

image (2)


src/modules/isLLVersionCompatible.js
```diff
@@ -0,0 +1,13 @@
+// TODO handle plugin version matrix in a better way
+export const minLLVersion = [100, 6, 0]
+
+export const isLLVersionCompatible = (version) => {
+    const versionArray = version.split('.').map((el) => parseInt(el, 10))
+
+    const isCompatible =
+        minLLVersion[0] <= versionArray[0] &&
+        minLLVersion[1] <= versionArray[1] &&
+        minLLVersion[2] <= versionArray[2]
```
KaiVandivier: It looks like v101.0.0 would return `false` here, so this logic needs a little adjustment -- here's a thought:

```js
const [minMajor, minMinor, minPatch] = minLLVersion
const [major, minor, patch ] = versionArray

isCompatible =
    major > minMajor ||
    major === minMajor && minor > minMinor ||
    major === minMajor && minor === minMinor && patch >= minPatch
```

KaiVandivier (changes requested):
I think the `isCompatible` logic needs a tweak and some more test conditions, but the rest looks good 👍

src/modules/isLLVersionCompatible.js
jenniferarnesen: Good catch. Fixed

KaiVandivier (approved):
Nice! Looks good to me 🙂

PR #2294, opened 2023-05-03
fix: plugin loading performance [patch/2.40.0]

jenniferarnesen:
This PR implements strategies to improve load time of the dashboard. The first time a dashboard is opened, only one plugin of each type (MAP, VISUALIZATION, LINE LIST) is loaded initially. Once those plugins report that they have loaded, then the other dashboard items proceed to load. This prevents multiple requests for the IFrame resources. 

In order to cleanly apply these changes to the patch/2.40.0 branch, 2 backports had to be included:

1. https://github.com/dhis2/dashboard-app/pull/2254
* the offline message was not showing on EE maps. A blank item was shown, leading user to think the map just wasn't loading.
* The legacy plugin (old ER and EV) was not resizing in edit mode.
* Refactor: no longer need DefaultPlugin (merged together with LegacyPlugin) and MapPlugin (everything now handled in Visualization.js).

2. https://github.com/dhis2/dashboard-app/pull/2272
* fix problem when switching between Map and DV using "View as"
Description
When using "View as" and switching between Map and chart/table or viceversa, the plugin loaded in the iframe needs to change.
This transition didn't work because the message listener in dashboard was not in place when the plugin requested getProps, resulting in the Map/chart/table not showing.
Switching between chart and table worked because in that case the plugin does not change and the newProps message is used instead of getProps.
The getProps listener is still removed in the cleanup function returned by the useEffect, but it's recreated if the iframe src change (plugin change).
After the first getProps message is received there shouldn't be other getProps messages incoming, but newProps is used for passing new/modified props instead.

About the changes in en.pot: The strings that appear to have been removed, were actually duplicated in the file, so this is a legit cleanup.


KaiVandivier (approved):
Tested the LL “version gate” and the “load first-of-type” features, and they seem to be working well 🙂 haven't tested the other things that were merged in though

edoardo (approved):

PR #2367, opened 2023-06-05
feat:  update gh workflows and set version to 100.0.0

jenniferarnesen:
BREAKING CHANGE: version will be bumped to v100.0.0, which is the base version for app-hub core apps.

The workflows are now updated so that PRs run against 2.40, and the nightly build runs against dev.


.github/workflows/dhis2-verify-app.yml
```diff
@@ -0,0 +1,174 @@
...
+                  group: 'e2e'
+                  tag: ${{ github.event_name }}
+              env:
+                  CI: true
+                  CYPRESS_RECORD_KEY: ${{ secrets.CYPRESS_RECORD_KEY }}
+                  COMMIT_INFO_MESSAGE: ${{ github.event.pull_request.title }}
+                  SERVER_START_CMD: 'yarn cypress:start'
+                  SERVER_URL: 'http://localhost:3000'
+                  cypress_dhis2_api_stub_mode: 'DISABLED'
+                  REACT_APP_DHIS2_BASE_URL: ${{ secrets.CYPRESS_DHIS2_BASE_URL_40 }}
```
edoardo, line 118: Is this wanted?
Does it mean we need to update the file whenever 41 is coming out?

edoardo (approved):

.github/workflows/dhis2-verify-app.yml
jenniferarnesen: This is only until we upgrade cypress. I tried reusing the shared workflows but there were cypress version issues. Right now, v40 is the only relevant prod version to test.

PR #2396, opened 2023-06-12
fix: release workflow should also run on patch/* branches

jenniferarnesen:
Pre 2.40 point releases of DHIS2 still rely on the app bundle in the d2-ci/dashboard-app repo, so the deploy-build step in this workflow must also run on patch/* 

patch/* should match for example (according to rules [gh action rules](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#filter-pattern-cheat-sheet))
patch/2.38.6
patch/2.39.3


jenniferarnesen:
Not needed

PR #2430, opened 2023-06-21
fix: use interpretations component from Analytics (DHIS2-15441)

jenniferarnesen:
Fixes: https://dhis2.atlassian.net/browse/DHIS2-15441
Depends on: https://github.com/dhis2/analytics/pull/1510
Depends on: https://github.com/dhis2/maps-app/pull/2936
Depends on: https://github.com/dhis2/maps-app/pull/2937

Replace the old d2-ui interpretations component with the one in analytics used by DV and LL. With this change, the old d2-ui sharing dialog is also replaced with the new one from ui.

Screenshots:
New:
image
Old:
image

Looking at one interpretation:
New:
image
Old:
image

Two old features were removed:
- `Show previous interpretations` button. This button toggled whether max 5 interpretations are shown, or all are shown.
image
- Replying to a reply. This is not even supported by the backend as all replies are connected directly with the original interpretation, so the reply button on a Reply never made sense anyway.
image

********************
**zip file size improvements:**
Old zip file: 5057221
New zip file: 3575577 (29% smaller)

**chunks (build/static/js/xyz.js)**
Old 
3053115 Jun 23 13:58 8877.6fc0d050.chunk.js

New:
1665003 Jun 23 13:55 396.371a423b.chunk.js (45% smaller)

Old source-map-explorer for dependencies bundle (all the dependencies in red were only there bc of the old interpretations component):
image

New source-map-explorer for dependencies bundle (all those red dependencies were removed):
image


HendrikThePendric (approved):
LGTM

PR #2431, opened 2023-06-21
refactor: use platform/runtime plugin components

edoardo:
Fixes [DHIS2-17268](https://dhis2.atlassian.net/browse/DHIS2-17268)

**Relates to https://github.com/dhis2/data-visualizer-app/pull/3082**
**Relates to https://github.com/dhis2/line-listing-app/pull/396**
**Relates to https://github.com/dhis2/maps-app/pull/3232**

---

### Key features

1. use plugin components from `app-runtime`
2. fix Cypress login for >=41

---

### Description

Generic components for plugins are available in `app-platform` and `app-runtime`.
This PR is removing the original iframe plugin with the message handling that has been "moved" to the platform.
All analytics apps have a correspondent PR for using the platform plugin build with embedded message handling.

Cypress started to fail in >=41 because the login mechanism changed.
Dashboard is still using v6.9.1 of Cypress which prevents us from upgrading `cli-utils-cypress` to a recent version with support for the new login.
The temporary solution is to add a custom login command which accommodates login for all DHIS2 versions.

---

### TODO

-   [x] figure out how to transmit the `removeCachedData` message to the plugin without re-rendering
-   [ ] handle `recordOnNextLoad` to avoid multiple recordings
-   [ ] test plugin type switch (ie. DV to Maps)
-   [ ] test installation status stuff

---

### Known issues

-   [ ] an infinite loop happened and [this](https://github.com/dhis2/dashboard-app/commit/03059501e58e2940efa5a750dac099cb612e5c29) seems to solve it, I'm not sure why and if this is a proper solution.

---

### Screenshots
Screenshot 2023-06-21 at 16 25 33

[DHIS2-17268]: https://dhis2.atlassian.net/browse/DHIS2-17268?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/pages/edit/ItemSelector/__tests__/__snapshots__/ContentMenuItem.spec.js.snap
```diff
@@ -3,14 +3,17 @@
 exports[`ContentMenuItem has a LaunchLink when url is provided 1`] = `
 <div>
   <li
-    class="jsx-1930534478 "
+    class="jsx-3192368000 "
```
jenniferarnesen, line 6: Maybe this test needs to be changed so we don't get these diffs when the ui version is upgraded.

jenniferarnesen:
Superceded by https://github.com/dhis2/dashboard-app/pull/3184

PR #2691, opened 2023-09-05
fix: parse the ou filter correctly for all ou types

jenniferarnesen:
Fixes [DHIS2-14544](https://dhis2.atlassian.net/browse/DHIS2-14544)

If path exists, then parse it, otherwise use id (which will be the case for groups, levels and USER_ORGUNITS)

[DHIS2-14544]: https://dhis2.atlassian.net/browse/DHIS2-14544?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/components/Item/AppItem/getIframeSrc.js
```diff
@@ -0,0 +1,15 @@
+import { FILTER_ORG_UNIT } from '../../../actions/itemFilters.js'
+
+export const getIframeSrc = (appDetails, item, itemFilters) => {
+    let iframeSrc = `${appDetails.launchUrl}?dashboardItemId=${item.id}`
+
+    if (itemFilters[FILTER_ORG_UNIT] && itemFilters[FILTER_ORG_UNIT].length) {
+        const ouIds = itemFilters[FILTER_ORG_UNIT].map(({ id, path }) =>
```
jenniferarnesen, line 7: @edoardo From what I've seen, all org units regardless if they are LEVEl, GROUP, individual from the hierarchy, or USER_ORG_UNIT* have the `id` property, so it seems unnecessary to parse the `path` at all. Do you remember if there was a reason to parse the path instead of use id?

edoardo: No, I can't remember, indeed if `id` was always present in `itemFilters` the parsing seems unnecessary.

src/components/Item/AppItem/getIframeSrc.js
```diff
@@ -0,0 +1,17 @@
+import { FILTER_ORG_UNIT } from '../../../actions/itemFilters.js'
+
+export const getIframeSrc = (appDetails, item, itemFilters) => {
+    let iframeSrc = `${appDetails.launchUrl}?dashboardItemId=${item.id}`
+
+    console.log('jj itemfilters', itemFilters)
```
edoardo: Remove the console log.

edoardo (approved):

PR #2801, opened 2023-10-23
fix: merge translations from past 3 weeks

jenniferarnesen:
In this PR:
* translations (makes this merge a 'fix' rather than a 'chore')
* fixes to cypress due to inconsistent return status codes from api (getting both 200 and 201)
* clean up cypress folder
* adds a release-failure slackbot message


janhenrikoverland (approved):
Approved if tests pass.

PR #2855, opened 2023-12-12
chore: include patch packages in cache-key

jenniferarnesen:
Related to https://dhis2.atlassian.net/browse/LIBS-383

Add package patches to the yarn cache key.

If a new patch is added, then this ensures that that the yarn cache gets invalidated and forces yarn install, which will then do the postinstall step that applies the patches.

From user perspective, the e2e-prod tests might fail if it is depending on a patch that didn't get installed.


KaiVandivier (approved):
Looks good to me 🙂

janhenrikoverland (approved):

PR #2917, opened 2024-02-26
test: adapt tests to new database 

martinkrulltott:
Just a dummy commit so far to run the CI tests, if all green no action is needed


martinkrulltott:
All tests are green so no Cypress updates are needed

PR #2970, opened 2024-05-03
fix: avoid reading from undefined error (DHIS2-17334)

edoardo:
Fixes [DHIS2-17334](https://dhis2.atlassian.net/browse/DHIS2-17334)

---

### Key features

1. fix read from undefined crash

---

### Description

This happened when using "View as" from Map to DV with some filters applied.
The reason is that `item` in that case is the Map item, and does not have the `visualization` key.
The tag part only make sense when rendering an Outlier table visualization type.
The fix makes sure the `item` is of type `VISUALIZATION` and then checks if the visualization type is an Outlier table one.

---

### Screenshots

Before:
Screenshot 2024-05-03 at 15 26 00

After:
the leftmost item is a Map one, and can be viewed as Chart even with filters applied
Screenshot 2024-05-03 at 15 26 26

the rightmost item is a Visualization of type Outlier table and should render a tag informing about some filters not applied (Area in the example)
Screenshot 2024-05-03 at 15 26 38

[DHIS2-17334]: https://dhis2.atlassian.net/browse/DHIS2-17334?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/Item/VisualizationItem/Item.js
+++ b/src/components/Item/VisualizationItem/Item.js
@@ -232,8 +232,9 @@ class Item extends Component {
                     }
                     case CHART:
                     case VISUALIZATION: {
-                        return item.visualization.type ===
-                            VIS_TYPE_OUTLIER_TABLE &&
+                        return item.type === VISUALIZATION &&
+                            item.visualization.type ===
+                                VIS_TYPE_OUTLIER_TABLE &&
                             Object.keys(itemFilters).some(
                                 (filter) =>
                                     ![


src/components/Item/VisualizationItem/Item.js
```diff
@@ -232,8 +232,9 @@ class Item extends Component {
                     }
                     case CHART:
                     case VISUALIZATION: {
-                        return item.visualization.type ===
-                            VIS_TYPE_OUTLIER_TABLE &&
+                        return item.type === VISUALIZATION &&
```
jenniferarnesen, line 235: Is there any use for the `originalType` here? Check the function `getItemTypeForVis`

jenniferarnesen (approved):

PR #3051, opened 2024-08-09
feat: support 3rd party dashboard plugins (DHIS2-17283)

edoardo:
Implements [DHIS2-17283](https://dhis2.atlassian.net/browse/DHIS2-17283)

---

### Key features

1. add support for 3rd party dashboard plugins

---

### Description

Dashboard app should have support for custom plugins built with the DHIS2 tooling.
This assumes the plugin uses the platform's components that provide the postRobot communication compatible with this version of dashboard app.

---

### TODO

-   [x] fix Jest tests for AppItem
-   [ ] update min versions for analytics apps

---

### Known issues

-   [ ] tests pass but for AppItem an error happens on render (the updated snapshot makes the test pass)

---

### Screenshots

Example with the Climate data app:
Screenshot 2024-09-06 at 16 21 59

[DHIS2-17283]: https://dhis2.atlassian.net/browse/DHIS2-17283?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/components/Item/ItemMessage/MissingPluginMessage.js
```diff
@@ -0,0 +1,45 @@
...
+import classes from './styles/ItemMessage.module.css'
+
+const MissingPluginMessage = ({ dashboardMode, pluginName }) => {
+    return (
+        <Cover>
+            <div className={classes.messageContent}>
+                <IconWarning24 color={colors.grey500} />
+                <span>
+                    {i18n.t(
+                        'The plugin for rendering this item is not available'
```
jenniferarnesen: Should we use a less technical word than "rendering" ?  Actually, is it even useful to mention that a plugin is missing? Not sure about the second question. If we decide not say "plugin", we could still put that info in a console.warning.

"The plugin needed to display this item is not available."
"This item can't be displayed. Please contact your administrator"

edoardo: I changed the message to your first suggestion.
As for the usefulness of the message, it does make sense for apps that are not bundled (ie. LL) and you open a dashboard which includes a visualization that needs a particular plugin.
In the same message we link to the App Hub and mention the name of the relevant app/plugin.

jenniferarnesen:
Superseded by https://github.com/dhis2/dashboard-app/pull/3184

PR #3056, opened 2024-08-16
chore: stop fetching schemas and replace d2 with dataEngine

janhenrikoverland:
The schemas aren't needed anymore.
Also, replaced d2 with dataEngine:
* post dataStatistics
* getting currentUser
* getting list of installed apps

We have to unfortunately disable some checks in the dashboard filter tests because cypress seems to be having a fit about the iframes. You could argue that the tests shouldn't be testing the content of the plugins anyway since we don't have control over which versions of the plugins are installed. The plugins should be getting tested in their respective apps. Still, it is sad to have to disable the checks.


janhenrikoverland (commented):
@jenniferarnesen will approve for @janhenrikoverland.

jenniferarnesen (approved):

PR #3071, opened 2024-08-27
fix: event chart and event report interpretations was showing an error

jenniferarnesen:
Fixes: https://dhis2.atlassian.net/browse/DHIS2-17943

Depends on: https://github.com/dhis2/analytics/pull/1697

Fixed: 

image
image


edoardo (approved):
Approved once the analytics dependency is bumped to the new published version.

PR #3081, opened 2024-09-06
feat: dashboard slideshow

jenniferarnesen:
Implements [DHIS2-13038](https://dhis2.atlassian.net/browse/DHIS2-13038)

Details about how it works:

1. When entering the Slideshow/fullscreen, the parent div of the ItemGrid is put into browser fullscreen. That same div remains in fullscreen the entire time, while the dashboard items are displayed/hidden in turn using css `opacity`. Css `opacity` is used instead of `display` or `visibility` so that the items can first be resized to fullscreen while "invisible", and then be made visible.

2. When first opening the slideshow, a flag (isPreSlideshow) is used to initially hide the item with opacity: 0 so that it can be resized and then appear fullscreen without the user seeing the resizing happen. This works ok, but I'm not sure it's worth the trouble

3. Progressive loading. When the dashboard goes into fullscreen/slideshow, then progressive loading, instead of being based on proximity to the viewport, will load the next and previous items in the fullscreen display order.

### TODO

-   [x] progressive loading
-   [x] check all item types
-   [x] small screen view (no fullscreen in small screen view)
-   [x] applied filters and active types
-   [x] verify print views, especially tables
-   [x] cypress tests
-   [x] Tested on large displays
-   [x] Disable nav buttons when there is only 1 item
-   [x] "Filter applied" label
-   [x] Offline should be possible to see visualizations if they are already loaded
-   [x] Hi-res screen - navigation bar is visibly resizing
-   [x] Opening legend on chart, lose keyboard navigation - this is expected behavior

[DHIS2-13038]: https://dhis2.atlassian.net/browse/DHIS2-13038?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


cypress/e2e/slideshow.feature
```diff
@@ -0,0 +1,20 @@
...
+        When I click the exit slideshow button
+        Then the normal view is shown
+
+
+    Scenario: I view fullscreen on the second item of the dashboard
+        Given I open the "Delivery" dashboard
+        When I click the fullscreen button on the second item
+        Then item 2 is shown in fullscreen
+        When I click the exit slideshow button
+        Then the normal view is shown
```
HendrikThePendric, line 20: Without having seen the actual implementation yet, I wonder if perhaps it would add value to add a test scenario for opening the last item and navigating backwards?

src/pages/view/styles/ItemGrid.module.css
```diff
@@ -1,3 +1,32 @@
...
+/* react-grid-layout uses width and height instead
+   off css logical properties, so these are the properties
+   that need to be overridden */
+/* stylelint-disable csstools/use-logical */
+.fullscreenItem {
+    width: 100% !important;
+    height: calc(100vh - 40px) !important;
+    transform: none !important;
+}
+/* stylelint-enable csstools/use-logical */
```
HendrikThePendric: Luckily I don't think this could cause any issue at all, since the item covers pretty much the entire screen 🤣

src/pages/view/useSlideshow.js
```diff
@@ -0,0 +1,105 @@
...
+    const [isPreSlideshow, setIsPreSlideshow] = useState(false)
+
+    // Sort items into order on dashboard
+    // and filter out items that don't support fullscreen
+    useEffect(() => {
+        const sItems = sortBy(displayItems, ['y', 'x']).filter((i) =>
+            itemTypeSupportsFullscreen(i.type)
+        )
+        sortedItems.current = sItems
+    }, [displayItems])
```
HendrikThePendric: Instead of using `useEffect`, I would recommend using `useMemo` here. 
```js
const sortedItems = useMemo(() => sortBy(displayItems, ['y', 'x']).filter((i) =>
    itemTypeSupportsFullscreen(i.type)
), [displayItems])
```

src/pages/view/useSlideshow.js
```diff
@@ -0,0 +1,105 @@
+import sortBy from 'lodash/sortBy.js'
+import { useState, useEffect, useRef, useCallback } from 'react'
+import { useSelector, useDispatch } from 'react-redux'
+import { acSetSlideshow } from '../../actions/slideshow.js'
+import { itemTypeSupportsFullscreen } from '../../modules/itemTypes.js'
+import { sGetSlideshow } from '../../reducers/slideshow.js'
+
+const useSlideshow = (displayItems, slideshowElementRef) => {
+    const dispatch = useDispatch()
+    const sortedItems = useRef([])
```
HendrikThePendric: This const should be replaces by the memoized const.

HendrikThePendric (changes requested):
These changes are looking great 😍 . I've left a few comments in the code, but nothing critical. I also tested the functionality briefly and did encounter an issue there. When navigating from one item to the next the black area at the bottom of the page seems to briefly expand. Is there anything that can be done about that.

Here's a screen recording:
https://github.com/user-attachments/assets/59bcb161-dde7-4ee2-96dd-146d77cb9c4d

Some additional info:
I'm on MacOS, using a MacBook Pro with M1 Max and showing the slideshow on an external 4K 32" monitor. My browser is Chrome Version 131.0.6778.109 (Official Build) (arm64).

**UPDATE**
I've also tested my on my laptop monitor and the issue cannot be reproduced there. I've also tried on that monitor with CPU throttling enabled, and still wasn't able to replicate the issue.... So it does not appear to be a performance related issue.....

**UPDATE 2**
I also tried scaling down the resolution of the external monitor and the problem goes away there too. So it seems to be related to screen resolution... The 2560x1440px is still working OK, but when I scale up to the next resolution 3008x1692px the problem starts to occur. Also, the amount by which the bottom bar increases in size during navigation seems be directly related to the resolution. On 3008x1692px I see a small height-increase, and when I scale the resolution up, each step gets a bigger height increase.

cypress/e2e/slideshow.feature
jenniferarnesen: Ill add a test for that

jenniferarnesen:
> These changes are looking great 😍 . I've left a few comments in the code, but nothing critical. I also tested the functionality briefly and did encounter an issue there. When navigating from one item to the next the black area at the bottom of the page seems to briefly expand. Is there anything that can be done about that.

Could you share a video? I don't see it.

HendrikThePendric (approved):
After fixing the large-monitor-issue and changing to `useMemo`, this PR looks great. Approved 🎉

cypress/e2e/slideshow.feature
jenniferarnesen: Done

janhenrikoverland (approved):

PR #3127, opened 2024-10-21
chore(app-platform): upgrade platform tools to use vite and react 18

kabaros:
[DHIS2-18794](https://dhis2.atlassian.net/browse/DHIS2-18794)

Relates to https://github.com/dhis2/app-platform/pull/887.

[DHIS2-18794]: https://dhis2.atlassian.net/browse/DHIS2-18794?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

Additional changes are:
* fix SonarQube errors

Quality checklist

- [ ] Docs added N/A
- [x] d2-ci dependency replaced (requires https://github.com/dhis2/analytics/pull/1785)
- [x] Small screen tested
- [x] Offline tested
- [x] Tester approved (@HendrikThePendric)


cypress.config.js
```diff
@@ -30,7 +30,7 @@ module.exports = defineConfig({
         specPattern: ['cypress/e2e_cucumber/*.feature', 'cypress/e2e/*.cy.js'],
         viewportWidth: 1280,
         viewportHeight: 800,
-        defaultCommandTimeout: 45000,
+        defaultCommandTimeout: 15000,
```
HendrikThePendric, line 33: I think it probably makes sense to use a reduced timeout. 45 seconds is an extremely long time to wait for a UI to update. But perhaps we should monitor E2E flakiness a bit more closely for a while.

package.json
```diff
@@ -5,17 +5,16 @@
     "private": true,
     "license": "BSD-3-Clause",
     "dependencies": {
-        "@dhis2/analytics": "^28.0.3",
-        "@dhis2/app-runtime": "^3.13.2",
+        "@dhis2/analytics": "git+https://github.com/d2-ci/analytics.git#d6e820515ef93e71e8b6e364e966a843a7d53efb",
```
HendrikThePendric: This still needs to be updated after we've tested things out

src/AppWrapper.jsx
```diff
@@ -1,13 +1,17 @@
...
+import AppDataProvider from './components/AppDataProvider/AppDataProvider.jsx'
+import WindowDimensionsProvider from './components/WindowDimensionsProvider.jsx'
 import configureStore from './configureStore.js'
 
 import './locales/index.js'
 
+if (typeof global === 'undefined') {
+    window.global = window
+}
+
```
HendrikThePendric, line 17: I happen to know this is to prevent issues with the `d2-utilizr` lib, but am likely to forget. It would be good to add a comment here.

HendrikThePendric (commented):
This wasn't the easiest PR to review because it contains so many unrelated changes and a the majority of the changes seem to be mostly syntax-related. I gave it my best shot though and left a few comments. The ones about the analytics lib and window.global will need to be resolved before I can approve.

One thing made me a bit uneasy while reviewing:
- Some E2E files have been moved and have had changes
- As a result they look like removed and new files while reviewing
- There are also a lot of tests commented out. I know this was already the case before this PR as well

But given the above, I keep thinking "is it still the same tests that are disabled? Or did more tests have get disabled in this PR"? The only way to check would be to compare the removed files with the new files and this would be extremely time consuming. So I think I'm just gonna have to live with this uneasiness...

jenniferarnesen:
> This wasn't the easiest PR to review because it contains so many unrelated changes and a the majority of the changes seem to be mostly syntax-related. I gave it my best shot though and left a few comments. The ones about the analytics lib and window.global will need to be resolved before I can approve.
> 
> One thing made me a bit uneasy while reviewing:
> 
> * Some E2E files have been moved and have had changes
> * As a result they look like removed and new files while reviewing
> * There are also a lot of tests commented out. I know this was already the case before this PR as well
> 
> But given the above, I keep thinking "is it still the same tests that are disabled? Or did more tests have get disabled in this PR"? The only way to check would be to compare the removed files with the new files and this would be extremely time consuming. So I think I'm just gonna have to live with this uneasiness...

I understand this was confusing. I've pulled out the Cypress test changes

src/AppWrapper.jsx
jenniferarnesen: I'll add a comment. It fixes the "legacy" plugins for EV and ER, as global is used there (see plugin.js)

package.json
jenniferarnesen: Yes, there is a task for this in the PR.

cypress.config.js
jenniferarnesen: At one point I had increased it to 45000 to give time for plugins to load. But now that most of the checks within the plugin have been removed, this long timeout shouldn't be needed.

package.json
jenniferarnesen: Updated

HendrikThePendric (approved):
LGTM

PR #3155, opened 2024-11-26
feat: view mode dashboards-bar design changes [DHIS2-18441]

HendrikThePendric:
Implements [DHIS2-18441](https://dhis2.atlassian.net/browse/DHIS2-18441) according to [design doc](https://docs.google.com/document/d/1c8Ll1aLbFYwU8HYsyDlH1wk_QKTrUXwNanCiqayczOM)

---

### Key features

1. New design
2. New navigation menu
3. New keyboard navigation

---

### Description

Implements the new design according to the design doc. Some other changes are also included in this PR and these are listed below.

#### Changes to `ViewDashboard`
- Use the `useAlert` hook instead of rendering a `AlertStack`. This ensures the show/hide animation runs and that there won't be problems when we show this alert together with another alert
- The logic to load a dashboard has been simplified, I suspect this component still contained some logic that predates the inclusion of react-router. In any case, there was some logic present that accommodated "loading a new dashboard". But this scenario would never occur, because when the route changes the component is remounted.
- Instead of having a `getContent` function, we now have the `ViewDashboardContent` component.

#### Changes to `CacheableViewDashboard`
A `useEffect` hook was added to ensure that the selected dashboard is cleared from the redux store when the user  chooses "Close dashboard" in the "action menu" dropdown.

#### Changes to `FilterBadge`
The design was updated and some modifications were done to the behaviour as well:
- When online on a large screen it is possible to edit and/or remove a filter
- When online on a small screen (width =< 480px) it is not possible to edit a filter, but it can be removed. When attempting to edit a filter a tooltip shows to inform the user this is not possible.
- When offline filters cannot be edited or deleted. The filter-badge is still visible but the remove-icon is not rendered. When attempting to edit a filter a tooltip shows to inform the user this is not possible.

#### Major version upgrade of `@dhis2/ui` (v10)
Version 10 includes some improvements in terms of accessibility and keyboard navigation. To comply with the new guidelines in that regard a few small tweaks to components were needed. This involved adding things like `ariaLabel` props, and addressing some invalid DOM nesting (nested `<a>` tags).

#### Some tweaks to the unit-testing configuration
- React testing-library was configured to work with the `data-test` attribute we use at DHIS2 (i.e. the `dataTest` prop for components from `@dhis2/ui`). The default data-attribute used by React testing-library is `data-test-id`, and some tests needed to be adjusted to no longer use that.
- Some ESLint rules were added specifically for `spec.js` to remove the need to add ESLint ignore comments when making mock-components and other types of mocks.

---

### TODO

-   [x] KFMT
-   [x] Add e2e test for error state, i.e. dashboard items request returning a 500

---

### Known issues

-   Keyboard navigation inconsistencies: this PR includes a major version bump in `@dhis2/ui` which comes with some accessibility improvements including arrow-key keyboard navigation for menus. The navigation- and more-actions-menu in the new dashboards-bar are using these. But the filter-menu is using a custom menu-list and this can be used by keyboard by using the tab-key. This inconsistency will be tackled in a separate PR, see [DHIS2-18537](https://dhis2.atlassian.net/browse/DHIS2-18537).

---

### Screenshots

Screenshot 2024-12-10 at 16 55 19

[DHIS2-18441]: https://dhis2.atlassian.net/browse/DHIS2-18441?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[DHIS2-18537]: https://dhis2.atlassian.net/browse/DHIS2-18537?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/pages/edit/ItemSelector/CategorizedMenuGroup.js
```diff
@@ -47,6 +47,10 @@ const CategorizedMenuGroup = ({
         onChangeItemsLimit(type)
     }
 
+    const showHideMoreLabel = seeMore
```
jenniferarnesen: `showFewerMoreLabel` is a more accurate name

src/pages/edit/ItemSelector/ContentMenuItem.js
```diff
@@ -2,22 +2,25 @@ import { visTypeIcons } from '@dhis2/analytics'
...
-        <IconLaunch16 color={colors.grey700} />
-    </a>
-)
+const LaunchLink = ({ url }) => {
+    const handleClick = useCallback(() => {
+        window.open(url, '_blank', 'noopener noreferrer')
+    }, [url])
+
+    return (
+        <span
```
jenniferarnesen, line 15: Was this change to get rid of the invalid html warning?

src/pages/view/FilterBar/FilterBadge.js
```diff
@@ -1,74 +1,73 @@
...
 import { sGetSelectedId } from '../../../reducers/selected.js'
 import classes from './styles/FilterBadge.module.css'
 
-const FilterBadge = ({ dashboardId, filter, openFilterModal, onRemove }) => {
-    const { isConnected: online } = useDhis2ConnectionStatus()
-    const { isCached } = useCacheableSection(dashboardId)
+const getFilterValuesText = (values) =>
+    values.length === 1
+        ? values[0].name
+        : i18n.t('{{count}} selected', { count: values.length })
```
jenniferarnesen: I wonder if we should handle plurals here, even though singular and plural are the same in this case in English. Might it be different in other languages?

src/actions/dashboardsFilter.js
```diff
@@ -10,6 +10,10 @@ export const acSetDashboardsFilter = (value) => ({
     value,
 })
 
+/* TODO: Possibly this action can be removed if we keep a an input
```
jenniferarnesen: Can we go ahead with this?

jenniferarnesen:
Check the constants in App.css related to the control bar. It's likely that many of these are no longer in use and can be removed.

src/pages/edit/ItemSelector/ContentMenuItem.js
HendrikThePendric: It was indeed. The MenuItem render an `<a>` already.

src/pages/view/FilterBar/FilterBadge.js
HendrikThePendric: I don't think we need to, because this is in a ternary expression and we only use this `i18n.t()` when `values.length` is larger than `1`. So it's always plural.

src/actions/dashboardsFilter.js
HendrikThePendric: Thanks for flagging this up. I had forgotten about this. I tried removing it, but after doing so found a few broken tests. This action is also called from the App component, along with some other `clearXXState` type of actions. In the end I think it's better to keep it. I did delete the comment though.

HendrikThePendric:
> Check the constants in App.css related to the control bar. It's likely that many of these are no longer in use and can be removed.

Thanks for the tip. I've removed these now.

src/pages/edit/ItemSelector/ContentMenuItem.js
jenniferarnesen: The semantics is a little backwards imho. I wonder why the menu item is an `<a>` when it is not really a hyperlink. It is just an action. Whereas the link to open the visualization in another app is more correct as an `<a>`. Anyway, not much to do about it.

src/pages/view/FilterBar/FilterBadge.js
jenniferarnesen: Yes, but some languages have multiple plural forms depending on how many things are being counted.

cypress/e2e/view_errors/error_while_fetching_dashboard_details.js
```diff
@@ -0,0 +1,23 @@
+import { When, Then } from '@badeball/cypress-cucumber-preprocessor'
+
+When('I type a dashboard id in the browser url that fails to load', () => {
+    console.log('TEST RUNS')
```
jenniferarnesen: remove console.log

cypress/e2e/view_errors/item_chart_fails_to_render.js
```diff
@@ -41,7 +41,7 @@ Given('I open a dashboard with a chart that will fail', () => {
 When(
     'I apply a {string} filter of type {string}',
     (dimensionType, filterName) => {
-        cy.contains('Add filter').click()
+        cy.contains('Filter').click()
```
jenniferarnesen: Do you think this is specific enough to always result in the correct element?

HendrikThePendric: I suppose it could break at some point. But I'd be in favour of keeping it until that happens. I like using text-based selectors for e2e tests where possible. In theory, we could change the implementation a lot but as long as the button label doesn't change (as it did this time around) the e2e test could still pass....

I could change it to a data-test selector, but would also have to add the data-test prop to the local `DropdownButton` component (which is not a big deal).

src/pages/view/CacheableViewDashboard.js
```diff
@@ -15,29 +16,32 @@ import {
...
+    clearSelectedDashboard,
     dashboardsIsEmpty,
+    dashboardsLoaded,
+    id,
+    selectedId,
 }) => {
-    const [dashboardsBarExpanded, setDashboardsBarExpanded] = useState(false)
     const { currentUser } = useCachedDataQuery()
 
+    useEffect(() => {
```
jenniferarnesen, line 31: What user scenario does this address?

src/pages/view/ViewDashboard.js
```diff
@@ -1,11 +1,13 @@
-import { useDhis2ConnectionStatus, useDataEngine } from '@dhis2/app-runtime'
+import {
```
jenniferarnesen, line 1: It's difficult to understand what behavior has changed in this component. Could you add an explanation to the PR description?

src/pages/view/ViewDashboard.js
```diff
@@ -15,191 +17,138 @@ import {
...
-            >
-                <DashboardsBar
-                    expanded={controlbarExpanded}
-                    onExpandedChanged={onExpandedChanged}
+        <div
+            className={cx(classes.container, 'dashboard-scroll-container')}
+            data-test="outer-scroll-container"
+        >
+            <DashboardsBar />
+            <DashboardContainer covered={false}>
```
jenniferarnesen: Seems safe to remove the `covered` property. This is the only place it is passed and it's always false.

src/pages/view/CacheableViewDashboard.js
HendrikThePendric: Before this code was in place, when a user would choose "Close dashboard" in the "action menu" dropdown, the selected dashboard would remain in the redux state. And now that we have this effect in place the redux state is cleared. I'm not 100% sure if this fix was also required to address a problem in the UI. I seem to remember that in some cases we would see the wrong information message. But in any case it does address a problem in the REDUX state, so good to keep anyway.

src/pages/view/FilterBar/FilterBadge.js
HendrikThePendric: AH OK, in that case we do have to add something like that. Could you advise on a change, since I am not familiar with this?

src/pages/view/ViewDashboard.js
HendrikThePendric: OK, I've done this, but note that while I have updated line 138 above in the current PR, I have removed the prop from the `DashboardContainer` component itself on another branch to avoid merge conflicts later on, see https://github.com/dhis2/dashboard-app/pull/3165/commits/c4b8117e4f29f11e7b2d6fbedf2b04ef38de7f63

src/pages/view/ViewDashboard.js
HendrikThePendric: I've added some text to the PR description now. Will resolve this comment.

src/pages/view/CacheableViewDashboard.js
jenniferarnesen: Ok, can you please note this in the PR description as well? When this PR is merged you can copy all the relevant info in the PR description into the squash commit message.

cypress/e2e/view_errors/item_chart_fails_to_render.js
jenniferarnesen: I was thinking rather you could do something like:

`cy.get('button').containsExact('Filter').click()`. 

You can get `containsExact` from DV. Not sure it's in Dashboard

src/pages/view/ViewDashboard.js
jenniferarnesen: So you're not going to remove `covered` from DashboardContainer in this PR? Seems most relevant to this PR since `covered` was all about the expanded control bar.

cypress/e2e/view_errors/item_chart_fails_to_render.js
HendrikThePendric: Yes it's there. I've switch to that helper in [6686ed1](https://github.com/dhis2/dashboard-app/pull/3155/commits/6686ed16e49d14ed83876b89c5775f5ab8d057e3)

jenniferarnesen (approved):

PR #3165, opened 2024-12-16
feat: maximize use of the available screen space by reducing whitespace in the dashboard item grid

HendrikThePendric:
Implements [DHIS2-18441](https://dhis2.atlassian.net/browse/DHIS2-18441) (grid improvements)

Note that these changes are not included in the [design doc](https://docs.google.com/document/d/1c8Ll1aLbFYwU8HYsyDlH1wk_QKTrUXwNanCiqayczOM/edit?tab=t.0#heading=h.96j4r17p7zj5) or the JIRA ticket.

---

### Key features

1. Reduces whitespace in the dashboard item grid
2. Adjusts layout when a scrollbar appears

---

### Description

In this PR we adjust the overall grid styles a bit to allow dashboard-items to be a little bit closer together. Additionally we wanted to solve the problem that our grid layout didn't take the presence/absence of a scrollbar into account. Before we simply had reserved sufficient margin on the right so that things would still look OK and function correctly when a scrollbar was present. In this PR an additional container was added which does not scroll and thus will reduce in width when its parent gets a scrollbar. Because it has a `ResizeObserver` attached to the DOM element we can monitor when this happens and trigger a state change. We've done this using a Provider/Consumer and a hook so that descendants can easily access this state.

---

##### Before

before_with_scrollbar

##### After

after_with_scrollbar

##### Before

before_without_scrollbar

##### After

after_without_scrollbar

##### Before

before_small_screen

##### After

after_small_screen

[DHIS2-18441]: https://dhis2.atlassian.net/browse/DHIS2-18441?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/modules/gridUtil.js
```diff
@@ -14,18 +14,16 @@ import {
...
+export const GRID_ROW_HEIGHT_PX = 16
+export const MARGIN_PX = [4, 4]
 
 const SM_SCREEN_MIN_ITEM_GRID_HEIGHT = 13 // minimum of ~320px
 export const SM_SCREEN_GRID_COLUMNS = 1
-export const MARGIN_SM_PX = [0, 16]
+export const MARGIN_SM_PX = [4, 4]
 export const GRID_PADDING_PX = [0, 0]
-// sum of left+right padding of dashboard-wrapper (App.css)
-export const DASHBOARD_WRAPPER_LR_MARGIN_PX = 32
```
HendrikThePendric, line 25: Note that this constant could be removed because we now work with that `div.contentWrap` which already has some whitespace around it and will shrink when a scrollbar appears.

src/pages/view/styles/ItemGrid.module.css
```diff
@@ -4,7 +4,6 @@
 }
 .grid {
     margin-block-start: var(--spacers-dp8);
-    background-color: #f4f6f8;
```
jenniferarnesen, line 7: This was added as part of the slideshow. Have you double-checked that there's no interruption of the grid bg color when entering slideshow, or switching between items?

HendrikThePendric: Aha, I figured it could be removed because the container already has a `background-color`. We should probably re-add it. But we do need to check what the correct color is because in `App.css` we're now using `var(--grey-200)` which is `#f3f5f7`. I would expect these to match?

HendrikThePendric: FYI: this change was done by me, but due to the squash commit I need to go back to the original PR to work out when I did this. And hopefully work out why as well. If I can't figure it out then perhaps we should perhaps revert this grey200 back to `#f4f6f8` as well.

HendrikThePendric: Aha! It was a change made by @cooper-joe in https://github.com/dhis2/dashboard-app/commit/20513e046018f60825dfd2aaa092868924d3b7e0 so we need to keep it. And your background-color should then also be `var(--colors-grey200)`. Do you agree?

_I'll do a commit assuming that you do agree_

HendrikThePendric: I've pushed https://github.com/dhis2/dashboard-app/pull/3165/commits/68ecbc060217f06642ce5d89c1dc49969e22f38e which re-adds a `background-color` to the `.grid` class and, but set it to `grey200` instead of #f4f6f8. And it also adjusts all backgrounds to have this color. But am still awaiting a reply from @cooper-joe.

HendrikThePendric: I received confirmation from @cooper-joe. This PR is now good to go.

jenniferarnesen (approved):

PR #3182, opened 2025-01-14
chore: sonarqube fixes

jenniferarnesen:
SonarQube analysis reported issues on the latest code changes. Even though this is not a required check, we should comply when the suggestions are sensible. Here's a link to the findings: https://sonarcloud.io/summary/new_code?id=dhis2_dashboard-app&branch=master

Summary of changes and the quality that was flagged by Sonarqube: 
* ActionsBar: break up nested ternary (maintainability)
* InformationBlock.module.css and SlideshowControlbar.module.css: missing generic font, but since this is declared higher up in the html structure for all elements (*), font-family declarations can be removed all-together in the app (reliability)
* NavigationMenu.js: remove role="listitem" on `li` element since it is is implicit (reliability, maintainability)
* ContentMenuItem.js: non-interactive element needs key handler. Fix was to make it semantically correct and switch to an interactive element `button` (reliability, maintainability)
* StartScreen.module.css: remove shorthand style that followed a more specific style (reliability)
* spec files - use object spread instead of Object.assign (conventional, maintainability)

What needs to be tested:
* start screen: that the bullets on the list of recent dashboards looks the same
* edit view item selector: that the launch link of each visualization looks the same and still works
* that the slideshow button has the correct tooltip content 
* InformationBlock and Slideshow controller have the correct font
* Dashboard dropdown "No dashboards found" looks correct

### DIFF
--- a/src/components/DashboardsBar/InformationBlock/ActionsBar.js
+++ b/src/components/DashboardsBar/InformationBlock/ActionsBar.js
@@ -186,14 +186,16 @@ const ActionsBar = ({
         return <Redirect to={redirectUrl} />
     }
 
-    let slideshowTooltipContent = null
-    if (!hasSlideshowItems) {
-        slideshowTooltipContent = i18n.t(
-            'No dashboard items to show in slideshow'
-        )
-    } else if (offline && !isCached) {
-        slideshowTooltipContent = i18n.t('Not available offline')
+    const getSlideshowTooltipContent = () => {
+        if (!hasSlideshowItems) {
+            return i18n.t('No dashboard items to show in slideshow')
+        } else if (offline && !isCached) {
+            return i18n.t('Not available offline')
+        }
+        return null
     }
+
+    const slideshowTooltipContent = getSlideshowTooltipContent()
 
     return (
         <>
--- a/src/components/DashboardsBar/InformationBlock/ActionsBar.js
+++ b/src/components/DashboardsBar/InformationBlock/ActionsBar.js
@@ -186,11 +186,14 @@ const ActionsBar = ({
         return <Redirect to={redirectUrl} />
     }
 
-    const slideshowTooltipContent = !hasSlideshowItems
-        ? i18n.t('No dashboard items to show in slideshow')
-        : offline && !isCached
-        ? i18n.t('Not available offline')
-        : null
+    let slideshowTooltipContent = null
+    if (!hasSlideshowItems) {
+        slideshowTooltipContent = i18n.t(
+            'No dashboard items to show in slideshow'
+        )
+    } else if (offline && !isCached) {
+        slideshowTooltipContent = i18n.t('Not available offline')
+    }
 
     return (
         <>
--- a/src/pages/edit/ItemSelector/styles/ContentMenuItem.module.css
+++ b/src/pages/edit/ItemSelector/styles/ContentMenuItem.module.css
@@ -5,12 +5,11 @@
 }
 
 .launchLinkButton {
+    all: unset;
     display: flex;
     margin-inline-start: 8px;
     z-index: 10000;
-    border: none;
-    background-color: transparent;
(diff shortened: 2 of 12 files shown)


src/components/DashboardsBar/InformationBlock/ActionsBar.js
```diff
@@ -186,11 +186,14 @@ const ActionsBar = ({
...
-        ? i18n.t('Not available offline')
-        : null
+    let slideshowTooltipContent = null
+    if (!hasSlideshowItems) {
+        slideshowTooltipContent = i18n.t(
+            'No dashboard items to show in slideshow'
+        )
+    } else if (offline && !isCached) {
+        slideshowTooltipContent = i18n.t('Not available offline')
+    }
```
HendrikThePendric: Perhaps extracting this into a function would be more "functional" than using mutable variable?

src/pages/edit/ItemSelector/styles/ContentMenuItem.module.css
```diff
@@ -4,10 +4,13 @@
...
 }
 
-.launchLink {
+.launchLinkButton {
     display: flex;
     margin-inline-start: 8px;
     z-index: 10000;
+    border: none;
+    background-color: transparent;
+    padding: 0;
```
HendrikThePendric: We could perhaps consider using `all: unset;` here? Or are there some button styles we want to preserve?

HendrikThePendric (changes requested):
I left two suggestions in the code, but nothing critical. 

I also compared the Netlify-deployed app with the app on https://test.e2e.dhis2.org/analytics-41dev/ and did spot a small regression.

> edit view item selector: that the launch link of each visualization looks the same and still works

The functionality is fine, but the cursor style is now a bit odd (on MacOS at least). The launch-link has a default cursor while the menu item has a `cursor: pointer;`. On the test instance they both have `cursor: pointer;`. FYI: it'd be good to also show a background color change on hover as a visual aid for the user, but I guess that's out-of-scope for now...

~One other note: I did not run the branch locally and could not work out how to test the following point on Netlify:~
> ~Dashboard dropdown "No dashboards found" looks correct~

~But since I could not actually spot any potentially relevant file changes in the PR diff, I just assumed this point would be fine.~
**I have tested this point too and it looked good.**

src/pages/edit/ItemSelector/styles/ContentMenuItem.module.css
HendrikThePendric: Also see my general PR comment, we do need `cursor: pointer;` and perhaps a highlight.

jenniferarnesen: That works.

jenniferarnesen:
> I left two suggestions in the code, but nothing critical.
> 
> I also compared the Netlify-deployed app with the app on https://test.e2e.dhis2.org/analytics-41dev/ and did spot a small regression.
> 
> > edit view item selector: that the launch link of each visualization looks the same and still works
> 
> The functionality is fine, but the cursor style is now a bit odd (on MacOS at least). The launch-link has a default cursor while the menu item has a `cursor: pointer;`. On the test instance they both have `cursor: pointer;`. FYI: it'd be good to also show a background color change on hover as a visual aid for the user, but I guess that's out-of-scope for now...
> 
> ~One other note: I did not run the branch locally and could not work out how to test the following point on Netlify:~
> 
> > ~Dashboard dropdown "No dashboards found" looks correct~
> 
> ~But since I could not actually spot any potentially relevant file changes in the PR diff, I just assumed this point would be fine.~ **I have tested this point too and it looked good.**

Addressed. The button now uses cursor pointer on hover, and the color of the icon is darker on hover. Nice improvement if you ask me...

HendrikThePendric (approved):

PR #3183, opened 2025-01-15
feat: add filter period to iframe src for custom widgets

EdsonNhancale:
Implements [DHIS2-19186](https://dhis2.atlassian.net/browse/DHIS2-19186)
### Key features

1. Support for adding `pe` (period) as a query parameter in the iframe source URL.
2. Dynamically appends the `pe` parameter to the iframe source when the filter is applied.

---

### Description

This update ensures that the `pe` (period) filter is included in the iframe URL for dashboard items when the filter is applied. If the `pe` filter is present in `itemFilters`, the periods are concatenated using `,` as the delimiter and added as a `period` query parameter in the iframe source.

---

### TODO

-   [ ] Verify the behavior with different combinations of filters (FILTER_ORG_UNIT and FILTER_PE).

---

### TESTS

-   [x] Added tests to validate the behavior of the period filter with single and multiple values.
-   [ ] Ensured proper handling when both pe and ou filters are applied together.
-   [x] Included a test to confirm period filter is omitted when empty
-   [ ] Verified the generated iframeSrc matches the expected URL for all scenarios.

---

### Known issues

-   [ ] When using custom widgets installed on the dashboard, the widget cannot currently react to changes in the period (pe) filter. This limitation arises because the pe filter value isn't dynamically communicated to the widget. A follow-up issue may be required to address this behavior.

---

### Screenshots
403252875-57e89dd7-3737-4297-9a33-5b5ee5181e84

403267034-3fc7b907-5603-4a61-9131-48e13d97a7a9

---

**Example:**
If the `itemFilters` contains:
```javascript
{
    FILTER_ORG_UNIT: [{ id: 'ou1', path: '/root/ou1' }, { id: 'ou2' }],
    FILTER_PE: [{id:'THIS_MONTH'}, {id:'LAST_MONTH'}]
}



[DHIS2-19186]: https://dhis2.atlassian.net/browse/DHIS2-19186?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### CODE REVIEW

File: src/components/Item/AppItem/getIframeSrc.js [hunk 1]
```diff
@@ -11,5 +11,12 @@ export const getIframeSrc = (appDetails, item, itemFilters) => {
         iframeSrc += `&userOrgUnit=${ouIds.join(',')}`
     }
 
+    // Add period (pe) to iframeSrc
+    if (itemFilters[FILTER_PE] && itemFilters[FILTER_PE].length) {
+        const peValues = itemFilters[FILTER_PE].map(x => x.id).join(';');
```
KaiVandivier: To match the format of the org unit parameter, let's use a comma as delimeter:

```suggestion
        const peValues = itemFilters[FILTER_PE].map(x => x.id).join(',');
```

src/components/Item/AppItem/getIframeSrc.js
```diff
@@ -11,5 +11,12 @@ export const getIframeSrc = (appDetails, item, itemFilters) => {
...
+        iframeSrc += `&pe=${peValues}`;
```
KaiVandivier: To match the org unit parameter, let's spell out `period`:
```suggestion
        iframeSrc += `&period=${peValues}`;
```

src/components/Item/AppItem/getIframeSrc.js
```diff
@@ -11,5 +11,12 @@ export const getIframeSrc = (appDetails, item, itemFilters) => {
         iframeSrc += `&userOrgUnit=${ouIds.join(',')}`
     }
 
+    // Add period (pe) to iframeSrc
+    if (itemFilters[FILTER_PE] && itemFilters[FILTER_PE].length) {
```
KaiVandivier: I know that this is the same logic as the expression above, but [Sonarqube](https://sonarcloud.io/project/issues?sinceLeakPeriod=true&issueStatuses=OPEN%2CCONFIRMED&pullRequest=3183&id=dhis2_dashboard-app&open=AZRoa7daNcCCOAJdJDaq) recommends using an optional chaining expression here:
```suggestion
    if (itemFilters[FILTER_PE]?.length) {
```

KaiVandivier (commented):
Thanks for opening this Edson! 🙏  The logic looks good 👍 

There's some things we should do to clean up the PR before we ask analytics folks to review it:
1. Add a more descriptive PR name, using conventional commit style, like "feat: add filter period to iframe src for custom widgets"
2. In the description, some code block styling is wrapping the screenshots, so the images aren't visible -- move those screenshots out of the code block so they're easy to see
3. There are some erroneous changes in `en.pot` and `FilterSelector.js` -- remove those
4. I added some inline comments in the code -- if you use the GitHub interface to apply them, make sure to use commit messages that follow the conventional commit style (start with `chore:`/`fix:`/`refactor:`)

Thanks again for setting this up! Let me know if you have any questions 🙂

jenniferarnesen:
Thank you for submitting this pull request. Could you please have a look at the unit tests in file `__tests__/getIframeSrc.spec.js` and add relevant tests for the new code including:
* only pe filter (including cases with just 1 and >1 filters)
* pe filter and ou filter

jenniferarnesen:
Hi Edson,
There are lint failures. You can ignore the `lint-commit` error (unless you want to rebase and fix it that way) because we'll do a squash commit when the time comes and rewrite the commit message at that point. But the `lint` error has to be fixed.

EdsonNhancale:
> Hi Edson, There are lint failures. You can ignore the `lint-commit` error (unless you want to rebase and fix it that way) because we'll do a squash commit when the time comes and rewrite the commit message at that point. But the `lint` error has to be fixed.

edoardo:
Hi @EdsonNhancale , before we can merge this there is a conflict with `master` that needs to be resolved.
Could you please merge `master` into your branch and ensure this is the content of `src/components/Item/AppItem/getIframeSrc.js` after the resolution?
Thanks!


```
import { FILTER_ORG_UNIT, FILTER_PE } from '../../../actions/itemFilters.js'
import { getPluginOverrides } from '../../../modules/localStorage.js'

export const getIframeSrc = (item, itemFilters, appDetails = {}) => {
    if (appDetails.appType === 'APP') {
        // pluginOverrides might be in use during development on localhost
        const pluginOverrides = getPluginOverrides()

        if (pluginOverrides?.[appDetails.key]) {
            return pluginOverrides[appDetails.key]
        }

        return appDetails.pluginLaunchUrl
    } else {
        let iframeSrc = `${appDetails.launchUrl}?dashboardItemId=${item.id}`

        if (itemFilters[FILTER_ORG_UNIT]?.length) {
            const ouIds = itemFilters[FILTER_ORG_UNIT].map(({ id, path }) =>
                path ? path.split('/').slice(-1)[0] : id
            )
            iframeSrc += `&userOrgUnit=${ouIds.join(',')}`
        }

        // Add period (pe) to iframeSrc
        if (itemFilters[FILTER_PE]?.length) {
            const peValues = itemFilters[FILTER_PE].map((x) => x.id).join(',')
            iframeSrc += `&period=${peValues}`
        }

        return iframeSrc
    }
}
```

jenniferarnesen:
@EdsonNhancale Could you also confirm whether the `Known Issue` in the PR description can be checked off?

KaiVandivier:
Perhaps this case is covered by the new ability to add custom app plugins to the Dashboard now?

jenniferarnesen:
> Perhaps this case is covered by the new ability to add custom app plugins to the Dashboard now?

This would be adding period filtering support for the older pre-existing dashboard widgets, which are different than the new plugin architecture. Since the older widgets will probably be around for some time, we should support it.

jenniferarnesen:
Superceded by https://github.com/dhis2/dashboard-app/pull/3255

Thank you for your contribution https://github.com/EdsonNhancale

PR #3185, opened 2025-01-16
feat: create embedded superset dashboard

HendrikThePendric:
This PR add the create flow:
1. When the user clicks the add button, if the system supports superset dashboards, the create flow will start. If not, the user will be taken straight to the "new internal dashboard view"
2. The first step is to select the type of dashboard. Id the user selects the internal type, he/she is taken to the "new internal dashboard view". If the external type is chosen, the user goes to the modal form
3. The only mandatory field in the modal form is the superset embed UUID. If the title is left empty the string "Untitled dashboard" will be used as the dashboard name

https://github.com/user-attachments/assets/38d1efc4-a9ef-4630-8d52-263a0327c280


src/components/DashboardsBar/CreateDashboardButton/CreateSupersetEmbeddedDashboard.js
```diff
@@ -0,0 +1,130 @@
...
+                filters: {
+                    visible: values.showFilters,
+                    expanded: false,
+                },
+            },
+        },
+    }),
+}
+
+export const CreateSupersetEmbeddedDashboard = ({
```
jenniferarnesen: A more intuitive name imho would be something like `ConfigureEmbeddedDashboardModal`. (point being that this component is a modal). Or, `ConfigureSupersetDashboardModal`

I'll use this same comment to mention that the name of the folder doesn't really represent the breadth of it contents. Could it be named `createDashboard` (camelCase)? Very open to discussion.

src/modules/useSupersetEmbeddedDashboardFieldsState.js
```diff
@@ -0,0 +1,80 @@
...
+    [fieldNames.description]: '',
+    [fieldNames.supersetEmbedId]: '',
+    [fieldNames.showChartControls]: true,
+    [fieldNames.showFilters]: true,
+}
+
+const FIELD_CHANGE = 'FIELD_CHANGE'
+const SUPERSET_FIELD_BLUR = 'SUPERSET_FIELD_BLUR'
+const UUID_PATTERN =
+    /^[0-9a-f]{8}-[0-9a-f]{4}-[1-5][0-9a-f]{3}-[89ab][0-9a-f]{3}-[0-9a-f]{12}$/i
```
jenniferarnesen, line 24: There's not some tiny 3rd party library that could validate the uuid instead? If we intend to use our own like this, then there should be unit tests for it.

src/components/DashboardsBar/CreateDashboardButton/ChooseDashboardTypeModal.js
```diff
@@ -0,0 +1,106 @@
...
+                <ModalTitle>
+                    {i18n.t('New dashboard: choose type', {
+                        nsSeparator: '###',
+                    })}
+                </ModalTitle>
+                <ModalContent>
+                    <fieldset
+                        ref={autoFocus}
+                        className={styles.dashboardTypeRadioGroup}
+                        tabIndex={0}
```
jenniferarnesen: This is triggering a SonarQube failure.

src/modules/useSupersetEmbeddedDashboardFieldsState.js
HendrikThePendric: I've had a look and actually think just using this UUID might be preferable. Widely used NPM packages that provide UUID validation have a much broader scope than validation alone. Quite likely the package would allow the excess to be tree-shaken off during the build, so perhaps we don't need to worry about bundle size, but we'd still be adding a substantial dependency to the codebase which does come with security implications (supply chain attacks are more and more common).

When looking at packages that solely focus on validation, [the most widely used one](https://github.com/microsoft/uuid-validate) (I could find) is still in v0 while not having seen any updates for years.

So I propose we stick with this regex. It is suppose to support UUIDs V1-V5. I do agree we need to add unit tests, but I don't think these need to be very extensive.

I'll be adding tests at a later stage so for now I've added a comment

src/components/DashboardsBar/CreateDashboardButton/ChooseDashboardTypeModal.js
HendrikThePendric: We've discussed that setting focus on the first input on each modal would be a suitable solution to address both the SonarCube failure and maintaining a good UX. See https://github.com/dhis2/dashboard-app/pull/3185/commits/89b5f2cfc1847fdc71dc227f14a679ccb8436a9b and https://github.com/dhis2/dashboard-app/pull/3185/commits/e4adf3575afacd6a0dc1f58aced23167ef27f90b

PR #3193, opened 2025-01-21
feat: adjust dashboards bar for external dashboards

HendrikThePendric:
This PR updates the dashboards-bar to deal with embedded dashboards. The main features are:
- Show a tag for embedded dashboards
- Disable various buttons in the `ActionsBar`, see [design specs](https://docs.google.com/document/d/1POHfTL48VfEN7VyR7yxIhg4TD08FI4YQVzuwciF_mb8/edit?tab=t.0#heading=h.pvxbt0ldmbhq) for details
- Allow edit/delete of embedded dashboards.

I have added a bunch of comments in the codebase to clarify some decisions I made.

_Screenshot 1: The dashboards-bar shows a tag and various menu items have been disabled_
Screenshot 2025-01-23 at 09 43 18

_Screenshot 2: After clicking the "Edit" button a modal form is displayed:_
Screenshot 2025-01-23 at 09 44 37


src/components/ConfigureSupersetEmbeddedDashboardModal/CreateSupersetEmbeddedDashboard.js
```diff
@@ -46,7 +31,9 @@ export const CreateSupersetEmbeddedDashboard = ({
     const dispatch = useDispatch()
     const history = useHistory()
     const [loading, setLoading] = useState(false)
-    const [postDashboard, { error }] = useDataMutation(postDashboardQuery)
+    const [postDashboard, { error }] = useDataMutation(postDashboardQuery, {
+        onError: () => setLoading(false),
+    })
```
HendrikThePendric, line 36: This change is not directly related to the current work, but I realised that in case of an error we would end up both in the error and the loading state here.

src/components/ConfigureSupersetEmbeddedDashboardModal/UpdateSupersetEmbeddedDashboard.js
```diff
@@ -0,0 +1,219 @@
...
+    const [mutationLoading, setMutationLoading] = useState(false)
+    const handeMutationError = useCallback(() => setMutationLoading(false), [])
+    const [updateDashboard, { error: updateError }] = useDataMutation(
+        updateDashboardQuery,
+        { variables: { id }, onError: handeMutationError }
+    )
+    const [deleteDashboard, { error: deleteError }] = useDataMutation(
+        deleteDashboardQuery,
+        { variables: { id }, onError: handeMutationError }
+    )
```
HendrikThePendric: I had to unset the loading state in case of errors in quite a peculiar way. A try/catch in `handleSubmit` and `handleDelete` would have been my preferred solution, but it turns out that the error is caught int the `useDataMutation` hook, so this was the only remaining solution unless we would switch to using the `dataEngine` directly.

src/components/ConfigureSupersetEmbeddedDashboardModal/UpdateSupersetEmbeddedDashboard.js
```diff
@@ -0,0 +1,219 @@
...
+                title: dashboard.name,
+                code: dashboard.code,
+                description: dashboard.description,
+                supersetEmbedId: dashboard.embedded.id,
+                showChartControls:
+                    !dashboard.embedded.options.hideChartControls,
+                showFilters: dashboard.embedded.options.filters.visible,
+            })
+        }
+    }, [dashboard, resetStateWithNewValues])
```
HendrikThePendric: Because I need to fetch the dashboard first, I ended up adding a new method to the `useSupersetEmbeddedDashboardFieldsState` hook. 

I considered using a nested component instead, which would load the dashboard in the parent and show the fields in the child (which could then be populated at mount time). 

The reason I went with this option is because this allows us to show the fields when the modal opens and thus will prevent the modal height from changing. But in the end this file turned out to be over 200 lines long, so perhaps splitting things up would have been a better option.

Let me know what you think.

src/components/ConfigureSupersetEmbeddedDashboardModal/UpdateSupersetEmbeddedDashboard.js
```diff
@@ -0,0 +1,219 @@
...
+                        )}
+                        <Button
+                            disabled={mutationLoading}
+                            secondary={!queryError}
+                            primary={!!queryError}
+                            onClick={closeModal}
+                            type={queryError ? 'submit' : undefined}
+                        >
+                            {i18n.t('Cancel')}
+                        </Button>
```
HendrikThePendric: When the dashboard-details fail to load, the "Cancel" button becomes the primary submit button, so a user can hit ENTER to close.

src/components/ConfigureSupersetEmbeddedDashboardModal/UpdateSupersetEmbeddedDashboard.js
```diff
@@ -0,0 +1,219 @@
...
+                            <Button
+                                destructive
+                                disabled={mutationLoading || queryLoading}
+                                secondary
+                                onClick={handleDelete}
+                                className={styles.deleteButton}
+                            >
+                                {i18n.t('Delete')}
+                            </Button>
+                        )}
```
HendrikThePendric: This is a little bit of lazyness on my side.... I should probably always show the button, but disable it with a tooltip explain that the user does not have access rights. Since this is not in the specs and because this modal shows after hitting an EDIT button I thought I could possibly get away with it.

src/components/ConfigureSupersetEmbeddedDashboardModal/UpdateSupersetEmbeddedDashboard.js
```diff
@@ -0,0 +1,219 @@
...
+                                primary
+                                disabled={
+                                    !hasFieldChanges ||
+                                    !isSupersetEmbedIdValid ||
+                                    queryLoading
+                                }
+                            >
+                                {i18n.t('Update dashboard')}
+                            </Button>
+                        )}
```
HendrikThePendric: Note that I am not checking on `dashboard.access.edit` here, or checking for small screens. this is because this modal only pops up after clicking the EDIT button in the dashboards-bar which already checks these things.

src/components/DropdownButton/DropdownButton.module.css
```diff
@@ -7,3 +7,7 @@
 .arrow {
     margin-inline-start: var(--spacers-dp12);
 }
+
+.arrow :global(path) {
+    fill: currentColor;
+}
```
HendrikThePendric, line 13: This ensures the icons gets the dimmed color when the button is disabled. This is a general improvement, but helps for the `FilterSelector` in particular.

src/components/ConfigureSupersetEmbeddedDashboardModal/UpdateSupersetEmbeddedDashboard.js
```diff
@@ -0,0 +1,219 @@
...
+                showFilters: dashboard.embedded.options.filters.visible,
+            })
+        }
+    }, [dashboard, resetStateWithNewValues])
+
+    return (
+        <Modal>
+            <form onSubmit={handleSubmit}>
+                <ModalTitle>
+                    {i18n.t('Edit external dashboard', { nsSeparator: '###' })}
```
jenniferarnesen: nsSeparator doesn't seem necessary here?

src/components/ConfigureSupersetEmbeddedDashboardModal/UpdateSupersetEmbeddedDashboard.js
HendrikThePendric: I decided to do different type of split. I've kept all the JSX in this component but have extracted the mutation logic into a hook. Quite happy with this new implementation.

src/components/ConfigureSupersetEmbeddedDashboardModal/UpdateSupersetEmbeddedDashboard.js
HendrikThePendric: Well spotted. Fixed [7436419](https://github.com/dhis2/dashboard-app/pull/3193/commits/7436419bd7cec0f61c4a69f819b30b111dd53567)

PR #3196, opened 2025-01-28
fix: force visualization fetch to ensure recording for offline

edoardo:
### Key features

1. ensure the visualization request is cached for offline

---

### Description

The visualization request stopped being recorded after recent changes. The componentDidUpdate method was taking care of the recording, but it doesn't work as before.
This change ensures the visualization is fetched when the dashboard is marked for offline so that the request is properly cached.


src/components/Item/VisualizationItem/Item.js
```diff
@@ -96,6 +96,11 @@ class Item extends Component {
                 await apiFetchVisualization(this.props.item)
             )
 
+        // force fetch when recording to allow caching of the visualizations request
+        if (this.props.isRecording) {
+            apiFetchVisualization(this.props.item)
```
jenniferarnesen, line 77: There is no need to await this request?

It does look a bit weird to have the same request 5 lines above.

Also, no chance that this request would get triggered twice?

edoardo: I don't think we need to await because we are not using the returned value.
The other fetch above needs the returned value for setting it in the Redux store.

I don't think it can be triggered twice because the recording can only happen once the dashboard is loaded.
But this needs testing.

PR #3197, opened 2025-01-29
test: add e2e tests for superset embedded dashboards feature

HendrikThePendric:
The main point of this PR is to add an e2e suite for the new embedded dashboards feature. To run this suite (locally) you will need have a v42 (current dev) DHIS2 Core backend in place and you will need to update the cypress environment config (`cypress.env.json`) so that version `42` of the web API is used. Note that you do **not** need a fully functioning superset and superset-gateway instance in place: we use `cy.intercept()` to fake that. 

While writing this PR some loosely related points were also addressed:
- It has been made possible to use the `excludeByVersionTags` plugin in this repo. Unfortunately we cannot use version tags on our tests yet because the `cypress-cucumber-preprocessor` and `cypress-tags` plugins cannot be used alongside each other. For now "version toggling" of e2e tests needs to be done using a hardcoded check, [like this one](https://github.com/dhis2/dashboard-app/blob/daee5c73bd34dd053fc10500f8dfd15af551c72c/cypress/e2e/embedded_superset_dashboard.cy.js#L29-L35).
- It has been made possible to run both cucumber based and "vanilla" cypress tests alongside each other. On of the steps to make this possible included moving separating the files into a `cypress/e2e` and `cypress/e2e_cucumber` dir. This will be convenient when we work on removing the cucumber tests in [DHIS2-18930](https://dhis2.atlassian.net/browse/DHIS2-18930)

Also while writing the tests, two bugs were identified and fixed.

Note that SonarCube has identified as many as 19 "new" issues, which are in fact all pre-existing issues which have now popped up because a lot of files have been moved from `cypress/e2e` to `cypress/e2e_cucumber`.

Also note that the `typescript` dev-dependency was added because it is needed by `cypress-tagify`. Additionally I also added `@babel/plugin-syntax-dynamic-import` for reasons I explain in https://github.com/dhis2/dashboard-app/pull/3197#discussion_r1937099169

The node-versions in the workflow files have also been upgraded. This was initially done to ensure the remote NodeJS environment matched my local setup where tests were passing, hoping that this would help getting the tests to pass on CI. Later on I realised that the e2e tests were failing on CI because a change was needed in `cypress/support/generateTestMatrix.js`, but I decided to keep the node version upgrades because it's a good bit of maintenance.

[DHIS2-18930]: https://dhis2.atlassian.net/browse/DHIS2-18930?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


package.json
```diff
@@ -42,7 +42,8 @@
         "cy:run": "start-server-and-test 'yarn cy:start' http://localhost:3000 'yarn cypress run --browser chrome headless'"
     },
     "devDependencies": {
-        "@badeball/cypress-cucumber-preprocessor": "^20.1.0",
+        "@babel/plugin-syntax-dynamic-import": "^7.8.3",
```
jenniferarnesen, line 45: Why is this dependency added?

package.json
```diff
@@ -65,7 +67,8 @@
         "redux-mock-store": "^1.5.4",
         "resize-observer-polyfill": "^1.5.1",
         "semantic-release": "^20",
-        "start-server-and-test": "^1.14.0"
+        "start-server-and-test": "^1.14.0",
+        "typescript": "^5.7.3"
```
jenniferarnesen, line 71: Could you add a comment in the PR description about why you added typescript?

package.json
HendrikThePendric: After I upgraded and added some some dev-dependencies, I cleared my `node_modules` and ran `npx yarn-deduplicate`. This resulted in a significant amount of changes in the `yarn.lock` and the package `@dhis2/cli-app-scripts` started complaining about this missing dependency, so I just added it. Since this was a change in the dev dependencies only I didn't worry about it too much. Do you think we need to investigate it further.

jenniferarnesen: No, I don't think so. But maybe just add a short comment to the PR description.

PR #3204, opened 2025-02-04
chore: remove d2, and Event Charts/Reports auth header used in development mode

jenniferarnesen:
The setting of the authentication header has caused a lot of headaches when testing
with non `admin` users as the app switches back to `admin` after login.
This is most likely only something happening during in-house development, but nevertheless,
the potential for it to impact actual installations is real if they were to have 
`REACT_APP_DHIS2_AUTHORIZATION` defined in their environment.

This header is no longer needed now that Event Charts and Event Reports use the `withCredentials`, which allows CORS requests.

Other changes:
* All remaining uses of d2 are now migrated to app-platform dataEngine.
** The `userDataStore` which is used when setting the show/hide description
** The utility function `generateUid` was copied from d2 into this repo.
** The api request fetchVisualization has been migrated.
* The show/hide description cypress test was reimplemented as a standard non-cucumber test, as that is where we are headed. Thus, we need side-by-side configurations for regular tests and cucumber tests. So, all the cucumber tests were moved to a `e2e_cucumber` subfolder, and non-cucumber config was added.

**Using localhost with Event Reports on analytics-dev:**
image

**Using netlify with Event Reports on analytics-dev:**
image

**In production (sanity check)**
image

**userDataStore requests to GET `dashboard`**
image

**userDataStore request to PUT `showDescription`**
image

**userDataStore requests to POST `showDescription`**
image

Confirmed that `visualization` requests are identical.


jenniferarnesen:
I think we're going to have to live with the SonarQube code smell report for now.

HendrikThePendric (approved):

HendrikThePendric (commented):
Code changes LGTM

HendrikThePendric (commented):
One thing I am considering for your PR and mine (https://github.com/dhis2/dashboard-app/pull/3205)....
Instead of having "vanilla cypress tests" in `cypress/e2e` and moving all of the existing stuff to `cypress/e2e_cucumber`, should we maybe keep all the old stuff where it is and move the new files to something like `cypress/e2e_plain`?
Maybe less disruptive that way?

jenniferarnesen:
> One thing I am considering for your PR and mine (#3205).... Instead of having "vanilla cypress tests" in `cypress/e2e` and moving all of the existing stuff to `cypress/e2e_cucumber`, should we maybe keep all the old stuff where it is and move the new files to something like `cypress/e2e_plain`? Maybe less disruptive that way?

Possibly. The only thing is that when we have migrated all the cucumber, then we'll probably want to rename `e2e_plain` to `e2e` which effectively shows the same changeset.

PR #3205, opened 2025-02-04
feat: embedded superset dashboards

HendrikThePendric:
### TODO list based on KFMT session

(remove this section once completed)

- [x] Documentation added
- [x] Fix: The warning text in the delete dialog accepts focus which should not be possible
- [x] Fix: the delete button in the delete dialog should auto-focus
- [x] Fix: modal should get scrolling content in case of limited screen height. 
- [x] Fix: Stop disabling the “save” button in the create/update modal and run validation before taking action, so that the user gets better feedback of when the UUID is not valid.
- [x] Show improved error messages and prevent flashing of embedded superset error page where possible. To do this we will request a guest token once before we embed the dashboard and if this fails we do not attempt to embed the dashboard but show an error notification instead. This task is currently blocked. The `superset-gateway/api/guestTokens` endpoint should start returning HTTP errors and error codes first.
- [x] In the sharing dialog we should prevent the second tab to show, since it is not possible to apply sharing settings to embedded dashboard items.
- [x] [STATUS: ✅  DONE] Change "External data" to "External source"
- [x] [STATUS: ✅  DONE] Change "Embed a dashboard from a third-party source, like Superset." to "Embed a superset dashboard"
- [x] [STATUS: ⛔ NOT POSSIBLE] Attempt to fix fullscreen option of individual embedded items by adjusting sandboxing properties
- [x] [STATUS: ⛔ NOT POSSIBLE] Attempt to fix "share by email" option of embedded item by adjusting sandboxing properties

---

Implements [DHIS2-17891](https://dhis2.atlassian.net/browse/DHIS2-17891)

### References

- [Embedded Dashboards](https://docs.google.com/document/d/13435SfkE8UQqWYkOiPtKlGMot0METPTuAfooZjDUR5k) entry point
- [Design document](https://docs.google.com/document/d/1POHfTL48VfEN7VyR7yxIhg4TD08FI4YQVzuwciF_mb8)
- [Workflow in the Dashboards app](https://docs.google.com/document/d/1XBAll5QaNjqFEzs7IRaXK169JrmnWAXKnpBK9ChDKi0/edit?tab=t.0#heading=h.kzgtwuiq321s)
- [Superset embedded SDK](https://www.npmjs.com/package/@superset-ui/embedded-sdk)
- [Jira epic](https://dhis2.atlassian.net/browse/DHIS2-18728)

---

### Key features

1. When the app loads it checks if the backend supports Superset embedded dashboards and enables them if this is the case
2. When creating a new dashboard, the user can choose to create an internal or external dashboard
3. Title, Code, Description and Superset Embed ID can be provided when creating/editing a Superset Embedded dashboard, as well as the options to "Show chart controls on dashboard items" and/or "Expand filters" in the emedded dashboard (using the Superset Embed SDK)
4. Embedded Superset dashboards can be starred, shared, and the description can be displayed/hidden, just like internal dashboards
5. They cannot, however, be filtered, taken offline, go into slideshow mode, or be printed. So these menu items are disabled

---

### Partial PRs

Several PRs have been merged into this main feature PR. Please review PR descriptions and commit messages for a more granular view on the changes:
- #3181
- #3185
- #3192
- #3193
- #3197
- #3206

---

### Note for testers

Since this feature is only available for instances which have Superset included, you will need to use such an instance when testing, i.e. https://debug.dhis2.org/dev. To test the dashboard-app, you can either use the Netlify link or run the app locally.

---

### Screenshots

_Choosing a dashboard type_
Screenshot 2025-02-10 at 12 47 49

_Configuring a Superset embedded dashboard_
Screenshot 2025-02-10 at 12 48 26

_Viewing a Superset embedded dashboard_
Screenshot 2025-02-10 at 12 49 00

[DHIS2-17891]: https://dhis2.atlassian.net/browse/DHIS2-17891?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/api/supersetGateway.js
```diff
@@ -0,0 +1,57 @@
+import { useConfig } from '@dhis2/app-service-config'
+import { useCallback, useMemo } from 'react'
+
+const useSupersetApiUrl = (resource) => {
```
jenniferarnesen: Perhaps each of these hooks could use a comment block with a little explanation?

src/components/DashboardsBar/NavigationMenu/NavigationMenuDropdownButton.js
```diff
@@ -0,0 +1,28 @@
+import i18n from '@dhis2/d2-i18n'
```
jenniferarnesen, line 1: I was a little confused about what this component was. I guess because I don't think of it as "navigating" to a dashboard. I think of this component more as a "DashboardSelector"

src/components/DashboardsBar/DashboardsBar.js
```diff
@@ -1,45 +1,15 @@
...
-                >
-                    <span className={styles.navMenuButtonText}>
-                        {i18n.t('Dashboards')}
-                    </span>
-                </DropdownButton>
-            </div>
-            <InformationBlock />
+export const DashboardsBar = () => (
+    <div className={styles.toolbar} data-test="title-bar">
+        <div className={styles.blockCreationNavigation}>
```
jenniferarnesen: Here the class name starts with "block", but on line 13, the component ends with "Block". Would it be more consistent to have this descriptor as a suffix?

src/components/Item/PrintTitlePageItem/Item.js
```diff
@@ -3,7 +3,7 @@ import PropTypes from 'prop-types'
 import React from 'react'
 import { connect } from 'react-redux'
 import { sGetIsEditing } from '../../../reducers/editDashboard.js'
-import { sGetNamedItemFilters } from '../../../reducers/itemFilters.js'
+import { msGetNamedItemFilters } from '../../../reducers/itemFilters.js'
```
jenniferarnesen, line 6: What is the "ms" convention? I haven't seen it before.

src/components/SystemSettingsProvider.js
```diff
@@ -56,3 +60,15 @@ SystemSettingsProvider.propTypes = {
 export default SystemSettingsProvider
 
 export const useSystemSettings = () => useContext(SystemSettingsCtx)
+export const useHasSupersetSupport = () => {
```
jenniferarnesen: `useSupersetIsSupported` perhaps?

src/modules/parseSupersetEmbeddedDashboardFieldValues.js
```diff
@@ -0,0 +1,17 @@
+export const parseSupersetEmbeddedDashboardFieldValues = (values) => ({
```
jenniferarnesen: Should `values` have a default {} ?

src/modules/useSupersetEmbeddedDashboardFieldsState.js
```diff
@@ -0,0 +1,94 @@
+import { useCallback, useReducer } from 'react'
+
+export const fieldNames = {
```
jenniferarnesen: Some years ago we decided that exported constants should not be contained in an object since it can be error prone. For example if the consumer misspells the prop name, it can go undetected.

Suggest instead to export each of these as a separate constant.

src/reducers/selected.js
```diff
@@ -37,6 +40,43 @@ export const sGetSelected = (state) => state.selected
...
+            filters: {
+                visible: true,
+                expanded: embedData.options.filters.expanded,
+            },
+        },
+    }
+}
+export const msGetSelectedSupersetEmbedData = createSelector(
+    [
+        (state) => state.selected?.embedded.id,
```
jenniferarnesen: If `embedded` can be undefined (line 17), could this crash?

src/components/DashboardsBar/NavigationMenu/NavigationMenuDropdownButton.js
HendrikThePendric: But to get to another dashboard you change the route by pushing to the `history`...
> It exposes useful methods and properties that let you **navigate** back and forth through the user's history

_(Source: [MDN docs on the History API](https://developer.mozilla.org/en-US/docs/Web/API/History_API), emphasis mine)_

I'm not really convinced that "selection" is more appropriate

src/components/DashboardsBar/DashboardsBar.js
HendrikThePendric: We can do, but this was a conscious decision on my side and you'll see this pattern all across the code I write, so perhaps good to explain my motivation before I blindly change things around.... My reasoning is as follows:
- In JavaScript, in general uses `camelCase`. Classes, component and global constants are exceptions.
- When using CSS Modules in a React/Webpack application, you can import a default variable from the CSS file. 
- I see the `styles` / `classes` import as an object and each class-name as a property of that object, and the value of that object property is the classname-string incl. the unique ID.

I don't know if the last step is technically true but it seems to be an accurate enough representation of things. And thus, if `styles` is a "regular js object" and `styles.className` is a "regular js property" then is should always be written in camelCase.

src/components/Item/PrintTitlePageItem/Item.js
HendrikThePendric: True it is a new convention I suggested and introduced for "memoized selectors". I discussed it with Jan the other day in [this Slack thread](https://dhis2.slack.com/archives/G3TL0J145/p1738768600819609). Let me know it the convention and the motivation behind it makes sense to you. And also if we need to document this somewhere.

src/modules/parseSupersetEmbeddedDashboardFieldValues.js
HendrikThePendric: I think not. There is no valid use-case for calling this function without a `values` argument. I'd rather have this function crash than it silently producing unexpected results.

(Looking forward to us switching to TypeScript, which will allow us to differentiate between required and optional parameters. Obviously we could then still have debates on wether a parameter should be optional with a default, or required, but at least my intentions would be 100% clear to you when you read the code)

src/components/DashboardsBar/DashboardsBar.js
jenniferarnesen: It's not about the camelCase. It's about where the word `block` is in the name. In the css class, `block` is the prefix. In the component, it is the suffix.

src/components/Item/PrintTitlePageItem/Item.js
jenniferarnesen: It's fine, just needed to know what it meant. Probably something we should share in some sort of conventions document or something.

src/components/SystemSettingsProvider.js
HendrikThePendric: Thanks for flagging this up. I think use `is` over `has` is preferable here as you suggest. In the end I went with `useIsSupersetSupported`, so that - when using this hook - a nice consistent `is`-prefixed boolean can be used:
```js
const isSupersetSupported = useIsSupersetSupported()
```

src/components/ConfigureSupersetEmbeddedDashboardModal/CreateSupersetEmbeddedDashboard.js
```diff
@@ -0,0 +1,118 @@
+import { useDataMutation } from '@dhis2/app-runtime'
```
jenniferarnesen: I'm thinking this component should be moved to a subcomponent of DashboardsBar. By convention that is how DashboardsBar components are organized. 

Also, I'm still hoping we can trim down this folder name and component names, and put "Modal" at the end of the Modal components (Create and Update)

src/components/DashboardsBar/NavigationMenu/NavigationMenuDropdownButton.js
jenniferarnesen: Fair point. But it's still a little misleading because there are other navigation elements. The "Edit" button, "New" button and "Print" menus. Maybe it could be called DashboardNavigator or DashboardNavigationMenuButton or similar. (Hard for me to say that, given I'm not a fan of really long names 😅 )

cypress.config.js
```diff
@@ -5,10 +5,14 @@ const createEsbuildPlugin = require('@badeball/cypress-cucumber-preprocessor/esb
 const createBundler = require('@bahmutov/cypress-esbuild-preprocessor')
 const { chromeAllowXSiteCookies } = require('@dhis2/cypress-plugins')
 const { defineConfig } = require('cypress')
+// const {
```
jenniferarnesen: We should just remove this commented code. At which point we need this, we can easily add it.

cypress/e2e/embedded_superset_dashboard.cy.js
```diff
@@ -0,0 +1,210 @@
...
+        // Can be starred via menu bar
+        cy.getByDataTest('dashboard-unstarred').click()
+        cy.getByDataTest('dashboard-starred').should('be.visible')
+        // And unstarred via ...menu
+        cy.getByDataTest('more-actions-button').should('be.enabled').click()
+        cy.contains('a', 'Unstar dashboard').should('be.visible').click()
+        cy.getByDataTest('dashboard-unstarred').should('be.visible')
+    })
+
+    it('can share the superset embedded dashboard', () => {
```
jenniferarnesen: rename test to reflect that it is not actually sharing. i.e. "can open the sharing dialog"

cypress/e2e/embedded_superset_dashboard.cy.js
```diff
@@ -0,0 +1,210 @@
...
+        cy.contains('h3', NAME).should('be.visible')
+        cy.contains('External data').should('be.visible')
+        // An iframe should be visible with the UUID in the src
+        cy.get('iframe')
+            .should('be.visible')
+            .and('have.attr', 'src')
+            .and('contain', UUID)
+    })
+
+    it('has some options disabled in the action-bar', () => {
```
jenniferarnesen: I think this test should be part of the previous test since it's just doing more confirmation that the dashboard was created. Also, I know we're already breaking Cypress best practice by having dependencies between tests, but the less of that we have the better.

cypress/e2e/embedded_superset_dashboard.cy.js
```diff
@@ -0,0 +1,210 @@
...
+            'true'
+        )
+        cy.contains('a', 'Print')
+            .should('have.attr', 'aria-disabled', 'true')
+            .and('have.attr', 'aria-expanded', 'false')
+        // Close the menu by clicking the backdrop
+        cy.get('.backdrop').should('be.visible').click()
+    })
+
+    it('shows the description', () => {
```
jenniferarnesen: I'm a little concerned about changing the show description and letting it linger. I get what you're doing, but we also have another test suite for testing show/hide description. If these test suites run concurrently, they could interfere with each other.

src/components/DashboardsBar/DashboardsBar.js
HendrikThePendric: Oh sorry I misread your comment 😊. I feel silly for writing such a lengthy reply that didn't make any sense. Note to self: spend more time reading to prevent wasting your own and other people's time 🤦 

Now that I have reread your comment, the suggestion makes perfect sense. Adding it to my todo list.

src/components/DashboardsBar/NavigationMenu/NavigationMenuDropdownButton.js
HendrikThePendric: As discussed on Slack: we're keeping this for now because even if the naming is confusing/inaccurate, at least it is consistent with its sibling components. We can consider a rename for this entire dir in a separate PR

cypress/e2e/embedded_superset_dashboard.cy.js
HendrikThePendric: Good point, I hadn't considered this.

src/components/DashboardsBar/InformationBlock/LastUpdatedTag.js
```diff
@@ -30,7 +30,7 @@ const LastUpdatedTag = ({ id }) => {
         >
             {(props) => (
                 <div {...props}>
-                    <Tag maxWidth="400px">{message} </Tag>
+                    <Tag maxWidth="200px">{message} </Tag>
```
jenniferarnesen: Is this changing the offline updated message? Won't 200px be too short?  
image

src/components/DashboardsBar/CreateDashboardButton/ChooseDashboardTypeModal.js
```diff
@@ -51,14 +45,9 @@ export const ChooseDashboardTypeModal = ({
...
-                            icon={<IconDashboardInternal />}
-                            title={i18n.t(
-                                'Internal: Data from {{systemName}}',
-                                {
-                                    systemName,
-                                    nsSeparator: '###',
-                                }
-                            )}
+                            title={i18n.t('Internal', {
+                                nsSeparator: '###',
```
jenniferarnesen: Is `nsSeparator` still needed (also lines 59-61)

src/components/DashboardsBar/ConfigureSupersetDashboard/styles/SupersetDashboardModal.module.css
```diff
@@ -33,6 +33,10 @@
     outline-offset: -3px !important;
     text-decoration: underline !important;
 }
+/* Prevent focus outline from shwoing on modal content */
```
jenniferarnesen: mispelling in comment

i18n/en.pot
```diff
@@ -5,14 +5,131 @@ msgstr ""
...
+msgid "Internal"
+msgstr "Internal"
+
+msgid "Show data and visualizations from this DHIS2 instance."
+msgstr "Show data and visualizations from this DHIS2 instance."
+
+msgid "External"
+msgstr "External"
+
+msgid "Embed a superset dashboard"
```
jenniferarnesen: Should "superset" be capitalized?

docs/dashboards.md
```diff
@@ -58,6 +58,8 @@ In create/edit mode, you can add or change the dashboard title, description, das
 
 ![](resources/images/dashboard-create-mode.png)
 
+In some DHIS2 Core instances it is possible to either create a regular (internal) dashboard, or to embed a dashboard from an external source. See [this section](#creating-and-editing-an-external-dashboard) for more information about this.
```
jenniferarnesen: I think this info should be emphasized as boxed text somehow, as an aside.

```
Note: if the system has been configured to support embedded dashboards, then you will have the option of creating a regular (internal) dashboard or embedding a dashboard from an external source. See [this section](#creating-and-editing-an-external-dashboard) for more information.
```

docs/dashboards.md
```diff
@@ -151,6 +153,27 @@ If you have access to delete the dashboard, then there will be a **Delete** butt
...
+If you choose the "External" option here and click "continue", you will be taken to a modal form where you can create an embedded dashboard:
+
+![Create external dashboard](resources/images/external-dashboard-create.png){ .center width=70% }
+
+The "Title", "Code" and "Description" fields in this form are identical to the fields with the same names you see when creating a regular (internal) dashboard, but there are also some notable differences:
+
+-   Since the dashboard content is coming from an external system (Superset), there is no need to add dashboard items
+-   A "Superset Embed ID" needs to be provided so the app can retrieve the correct dashboard from the external Superset service
+-   The "Show chart controls on dashboard items" checkbox controls whether or not each embedded dashboard item has a dropdown menu
+-   The "Expand filters" checkbox controls whether or not the filter section of the embedded dashboard is expanded or collapsed
```
jenniferarnesen: is initially expanded or collapsed

docs/dashboards.md
```diff
@@ -151,6 +153,27 @@ If you have access to delete the dashboard, then there will be a **Delete** butt
...
+
+![Superset dashboard with both checkboxes enabled](resources/images/superset-dashboard-with-annotations.png)
+
+Once the dashboard is created, you can star/unstar, edit and share it. You can also show and hide the description and close the dashboard. But it is not possible to show it in slideshow mode, filter it, or enter print view mode.
```
jenniferarnesen: Remove the "But"

jenniferarnesen (approved):

docs/dashboards.md
HendrikThePendric: We discussed this on Slack and concluded we would use the sentence @jenniferarnesen suggested but without the `Note:` prefix, because in the DHIS2 Core docs Notes are typically in emphasised blocks and we do not want to emphasise this too much, since it is a feature which is not often available.

PR #3209, opened 2025-02-07
fix: handle special characters in confirm delete dialog and no dashboards found message [v39]

jenniferarnesen:
Part of implementation for Implements [DHIS2-6647](https://dhis2.atlassian.net/browse/DHIS2-6647)

image

image

[DHIS2-6647]: https://dhis2.atlassian.net/browse/DHIS2-6647?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


HendrikThePendric (approved):
I think in both of these cases it should be safe to do this.

PR #3223, opened 2025-02-17
fix: post correct eventtype for dataStatistics requests

jenniferarnesen:
Fixes https://dhis2.atlassian.net/browse/DHIS2-17935 and https://dhis2.atlassian.net/browse/DHIS2-19022


janhenrikoverland (approved):
LGTM

PR #3226, opened 2025-02-21
feat: lazy load dashboards into the navigation menu [DHIS2-10624]

jenniferarnesen:
Implements [DHIS2-10624](https://dhis2.atlassian.net/browse/DHIS2-10624)

---

### Key features

To improve dashboard app performance for users with access to many dashboards, use paging to limit the number of dashboards loaded into the navigation menu

### Description

The navigation menu now has "infinite" scrolling similar to the data item selector in DV and LL. Scrolling to the bottom triggers a fetch of the next page of results (page size set to 200 as suggested by Lars in the Jira ticket). Searching works the same way, sending a request with the filter text and get the results, paged.  Searching is debounced to 300ms.

The CacheableViewDashboard code is primarily determining whether there is a dashboard the user can view: when the app initially loads, instead of fetching all the dashboards like before, it now fetches a single dashboard to answer the question of whether there are any dashboards.

The list of dashboards has been removed from redux and is now kept in the NavigationMenu state. Each time the user opens the menu, the first page of dashboards is fetched.

The "starred" status of the dashboard being displayed is now kept in the redux 'selected' property.

The component EndIntersectionDetector was adapted from the Data Item selector (transfer) component of the same name (from analytics).

---

### TODO

- [ ]  Test on instance with 1000s of dashboards (https://dev.im.dhis2.org/lazy-load-dashboards, user system)
- [ ]  Test on instance with 0 dashboards (https://dev.im.dhis2.org/no-dashboards-1, user admin)
- [ ] Offline is supported
- [ ] Regression test Superset dashboard
- [ ] Small screen
---

[DHIS2-10624]: https://dhis2.atlassian.net/browse/DHIS2-10624?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/components/App.js
```diff
@@ -27,21 +25,11 @@ import './styles/ItemGrid.css'
 const App = (props) => {
     const { systemSettings } = useSystemSettings()
     const { currentUser } = useCachedDataQuery()
+    const { setShowDescription } = props
 
     useEffect(() => {
-        props.fetchDashboards()
-        props.setShowDescription()
-
-        // store the headerbar height for controlbar height calculations
```
jenniferarnesen: This headerbar height is no longer relevant as of the new dashboard design

PR #3230, opened 2025-02-27
chore: add e2e dev workflow

HendrikThePendric:
I think we need to blindly merge this PR so we have the workflow on the main branch. We can then start new branch from master and add the `e2e dev` label to it. I believe we should then see the job running on each commit. And that means we can start tweaking things until they work.

There is also the off chance that I will have got everything right the first time around, but that would be quite unexpected 😄

### DIFF
--- a/.github/workflows/e2e-dev.yml
+++ b/.github/workflows/e2e-dev.yml
@@ -18,7 +18,7 @@ defaults:
 
 jobs:
     compute-dev-version:
-        if: contains(github.event.pull_request.labels.*.name, 'e2e-record')
+        if: contains(github.event.pull_request.labels.*.name, 'e2e dev')
         runs-on: ubuntu-latest
         outputs:
             version: ${{ steps.instance-version.outputs.version }}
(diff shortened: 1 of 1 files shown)


.github/workflows/e2e-dev.yml
```diff
@@ -0,0 +1,73 @@
...
+        steps:
+            - name: Output dev version
+              id: instance-version
+              uses: dhis2/action-instance-version@v1
+              with:
+                  instance-url: ${{ env.DEV_URL }}
+                  username: ${{ secrets.CYPRESS_DHIS2_USERNAME }}
+                  password: ${{ secrets.CYPRESS_DHIS2_PASSWORD }}
+
+    e2e-record:
```
jenniferarnesen: Should the job name instead be `e2e-tests-dev` or similar?

.github/workflows/e2e-dev.yml
```diff
@@ -0,0 +1,73 @@
...
+    group: ${{ github.workflow}}-${{ github.ref }}
+    cancel-in-progress: true
+
+defaults:
+    run:
+        shell: bash
+
+jobs:
+    compute-dev-version:
+        if: contains(github.event.pull_request.labels.*.name, 'e2e-record')
```
jenniferarnesen: Shouldn't we use a different label than e2e-record to determine whether to run this workflow (here and line 36)? Most of the time I imagine people just want to run the tests without recording since it isn't needed most of the time.

HendrikThePendric: Yeah sorry, this should have been `e2e dev`. Side note: I should also not have used hyphens in the label names.

jenniferarnesen (approved):

PR #3236, opened 2025-03-04
fix: ensure app version is returned for bundled apps [DHIS2-19138]

edoardo:
Implements [DHIS2-19138](https://dhis2.atlassian.net/browse/DHIS2-19138)

---

### Key features

1. lookup app version in the bundled apps list if nothing is returned from `api/apps`

---

### Description

For apps that are bundled and don't have a manually installed version, `api/apps` returns nothing.
This causes dashboard to think there isn't an installed version of the app/plugin and shows a warning in the dashboard item. The app version should be looked up in the bundled apps list before returning "no version" (0.0.0).

---

### TODO

---

### Screenshots

Before.
For bundled apps with no manually installed version, `api/apps` returns nothing.
A warning is shown in the dashboard item:
Screenshot 2025-03-04 at 17 04 33

After.
The app version is looked up first in `api/apps` and then in `dhis-web-apps/apps-bundle.json`:
Screenshot 2025-03-04 at 17 04 51

[DHIS2-19138]: https://dhis2.atlassian.net/browse/DHIS2-19138?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


jenniferarnesen (commented):
Looks good, except for the lint error that should be addressed.

HendrikThePendric (changes requested):
I have some general suggestions for refactoring this for readability and getting rid of the ESLint warning, but I have also spotted a "real" problem with the current implementation:
- The `appVersions` state is being initialised as an empty array
- The final `appVersions` state is computed with the results of a GET request
- The children are being rendered immediately, which means that while the GET request is pending, the descending components are not able to read the app-versions
- So ideally, I guess this component requires a loading state (and perhaps also an error state). But at the very least, it should render the children conditionally, i.e. `{appVersions.length > 0 ? children : null}`

Some more general suggestions:
- Consider consolidating the "all the app fetching logic" into one place. So this would mean removing the get for `apps` from the `CachedDataProvider` and moving it to the `InstalledAppVersionsProvider`*. You could use `useDataQuery` for this.
- Consider moving the fetch for bundled apps to dedicated hook: `useFetchBundledApps`, which mimics the `useDataQuery` behaviour (i.e. it should return `{ data, error, loading }`
- After implementing the suggestions above, it should be easy to clean up the provider component

[*] I realise that after doing this, we'll have to scan the full codebase and identify places where we read `apps` from the `useCachedDataQuery` hook, and then change it. But I don't think that would be too disruptive.

HendrikThePendric (approved):
Great work!

PR #3238, opened 2025-03-10
fix: support bundled LL app (DHIS2-17400)

edoardo:
Implements [DHIS2-17400](https://dhis2.atlassian.net/browse/DHIS2-17400)

---

### Key features

1. ensure LL plugin is loaded when LL is bundled

---

### Description

Since `api/apps` does not return bundled apps that don't have a manually installed version, a hardcoded path is needed for pointing the iframe to the correct plugin. This is the same approach we already use for DV and Maps.

This change shouldn't affect the existing way of loading the LL plugin as currently the `pluginLaunchUrl` from `api/apps` is always used if LL app is installed.
The manually installed versions always take precedence, the bundled apps hardcoded paths are used as fallback.

---

### TODO

- [x] Confirm the [app path](https://github.com/dhis2/dashboard-app/pull/3238/files#diff-8f42b6ed57ff11ed31fcc4ef778a5780343aa6e8e729d36531b439b02855b5dfR52) once the ticket is done
- [x] Test when a bundled LL app is installed on some instance

[DHIS2-17400]: https://dhis2.atlassian.net/browse/DHIS2-17400?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


janhenrikoverland:
@edoardo We probably need a 42 toggle?

edoardo:
> @edoardo We probably need a 42 toggle?

I don't think we need any toggle.
In 42 `api/apps` seems to return all apps, including bundled ones.
This means that the plugin URL used is read from there.
For the LL app, the fallback to the hardcoded path is never going to be used.
In <=41 we have the "fallback" where we use an "hardcoded" path for bundled apps based on the dashboard item type, but LL is not bundled there, so the plugin URL is anyway read from `api/apps`.

src/components/Item/VisualizationItem/Visualization/plugin.js
```diff
@@ -22,7 +23,9 @@ const itemTypeToScriptPath = {
 }
 
 const hasIntegratedPlugin = (type) =>
-    [CHART, REPORT_TABLE, VISUALIZATION, MAP].includes(type)
+    [CHART, REPORT_TABLE, VISUALIZATION, MAP, EVENT_VISUALIZATION].includes(
```
jenniferarnesen: Isn't this only the case (EVENT_VISUALIZATION) for 42?

edoardo: That's a good observation.
I don't think it matters for 42 because `api/apps` returns bundled apps too, so `hasIntegratedPlugin` is actually skipped in 42 for all bundled apps.
We should then revert this code so it works as it does currently in 40/41.

edoardo:
This is actually no longer needed because in 42 `api/apps` returns all apps, including bundled apps.

janhenrikoverland (approved):

PR #3245, opened 2025-03-19
fix: hide the global header in print mode [DHIS2-19241]

jenniferarnesen:
Implements [DHIS2-19241](https://dhis2.atlassian.net/browse/DHIS2-19241)

---

### Description

If the global shell header exists in the top document, then show/hide that header when in print mode. Otherwise, show/hide the current header.

Test instance for 42 with custom build deployed: https://dev.im.dhis2.org/jen-analytics-42-0

---

### TODO

- [ ] Docs added N/A
- [ ] i18n strings generated N/A
- [ ] d2-ci dependency replaced N/A
- [x] tested in 40, 41 (no global shell)
- [x] tested in 42 both with and without global shell

---

### Screenshots
Before:
image

After:

image

[DHIS2-19241]: https://dhis2.atlassian.net/browse/DHIS2-19241?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


cypress/e2e_cucumber/view_dashboard/print.js
BRaimbault, line 1: Does not look like: "the print one-item-per-page displays for {string} dashboard" is used anywhere else, is it? You may consider removing it too.

BRaimbault (dismissed):

BRaimbault (approved):

PR #3246, opened 2025-03-20
fix: ensure fields do not have undefined initial values

HendrikThePendric:
Implements [DHIS2-19266](https://dhis2.atlassian.net/browse/DHIS2-19266)

---

### Key features

1. When creating the initial state, ensure the `values` and `initialValues` do not contain properties that are `undefined`

---

### Description

The error we were seeing when opening the modal was caused by the fact that a dashboard did not have a code. So the `initialValues` object provided to `createInitialState` was not undefined itself, but one of its properties did have a value of `undefined`. The validation logic was assuming each property would default to the value defined in the `defaultInitialValues` object, but this was the case.

In this fix I chose to implement a generic fix that scans all `initialValues` properties and if an `undefined` value is encountered, it falls back to the default. IMO this preferable over making the validation logic more flexible, or implementing a fix specific to the `code` field. 

---

[DHIS2-19266]: https://dhis2.atlassian.net/browse/DHIS2-19266?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/package.json
+++ b/package.json
@@ -54,6 +54,7 @@
         "@semantic-release/git": "^10",
         "@testing-library/jest-dom": "^6.6.3",
         "@testing-library/react": "^12",
+        "@testing-library/react-hooks": "^8.0.1",
         "cypress": "^13.13.1",
         "cypress-real-events": "^1.13.0",
         "cypress-tags": "^1.2.2",
--- a/src/components/DashboardsBar/ConfigureSupersetDashboard/SupersetDashboardFields.js
+++ b/src/components/DashboardsBar/ConfigureSupersetDashboard/SupersetDashboardFields.js
@@ -25,6 +25,7 @@ export const SupersetDashboardFields = ({
             <InputField
                 initialFocus
                 label={i18n.t('Title')}
+                placeholder={i18n.t('Untitled dashboard')}
                 type="text"
                 onChange={onChange}
                 value={values.title}
--- a/src/modules/__tests__/useSupersetDashboardFieldsState.spec.js
+++ b/src/modules/__tests__/useSupersetDashboardFieldsState.spec.js
@@ -1,17 +1,10 @@
+import { renderHook, act } from '@testing-library/react-hooks'
 import {
-    createInitialState,
-    defaultInitialValues,
-    FIELD_CHANGE,
     FIELD_NAME_CODE,
-    FIELD_NAME_DESCRIPTION,
-    FIELD_NAME_EXPAND_FILTERS,
     FIELD_NAME_SHOW_CHART_CONTROLS,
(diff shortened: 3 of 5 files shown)


jenniferarnesen:
So, I'm a follower of the idea of only testing the intended public interface of a module (hook in this case) rather than testing implementation details (e.g. isValidUuid, createInitialState). This keeps code cleaner and prevents unintended uses of those "internal" functions. It also makes it easier to identify when a constant or function is not in use and to refactor the implementation. As far as testing goes, in the case of the current tests, it is really easy to do these tests just by testing the hook:

```
      it('creates the expected initial state when some properties are undefined', () => {
            const initialValues = {
                title: undefined,
                code: undefined,
                description: undefined,
                supersetEmbedId: undefined,
                showChartControls: undefined,
                expandFilters: undefined,
            }
            const { result } = renderHook(() =>
                useSupersetDashboardFieldsState(initialValues)
            )

            expect(result.current.values).toEqual(defaultInitialValues)
            expect(result.current.isSupersetEmbedIdValid).toBe(false)
            expect(result.current.isCodeValid).toBe(true)
            expect(result.current.shouldShowSupersetEmbedIdError).toBe(false)
            expect(result.current.hasFieldChanges).toBe(false)
      })
```

HendrikThePendric:
> So, I'm a follower of the idea of only testing the intended public interface of a module (hook in this case) rather than testing implementation details (e.g. isValidUuid, createInitialState). This keeps code cleaner and prevents unintended uses of those "internal" functions. It also makes it easier to identify when a constant or function is not in use and to refactor the implementation.

This is interesting. Up until now I have mostly given preference to testing the smallest unit in isolation when writing unit tests. And while I do think that in general this is a sensible notion, and that this will actually result in the cleanest code, I do see your other points and will start writing unit tests against the intended public interface in the future. My current approach does involve exporting these internals and I have seen the problems you mention happen in practice.

I'll proceed to update the entire spec file.

jenniferarnesen (changes requested):
Can you add the "Untitled dashboard" placeholder to the title, since that's what it gets set to if the user doesn't type anything. This is also in line with regular dashboard creation.

HendrikThePendric:
> Can you add the "Untitled dashboard" placeholder to the title, since that's what it gets set to if the user doesn't type anything. This is also in line with regular dashboard creation.

OK, done https://github.com/dhis2/dashboard-app/pull/3246/commits/0109115e15b4e15a12824cf5811472ece3f4302d

jenniferarnesen (approved):

PR #3247, opened 2025-03-20
fix: display error when Superset gateway configuration is invalid [DHIS2-19275]

HendrikThePendric:
Implements [DHIS2-19275](https://dhis2.atlassian.net/browse/DHIS2-19275)

---

### Key features

1. App does not crash when Superset Gateway is not available
2. All other functionality remains identical. This means users can still see all embedded dashboards in the list and navigate to them. They can also create new ones and view and edit them.
3. However, instead of actually showing any dashboard content, the user will be presented with the standard error UI, with a specific message.

---

### Description

Apart from the points mentioned in the key features, this PR also addresses a bug [3c5fe12](https://github.com/dhis2/dashboard-app/pull/3247/commits/3c5fe1238e0b1352fadab0fafd012413fb4c60bd) and removes a redundant selector function [924d4c8](https://github.com/dhis2/dashboard-app/pull/3247/commits/924d4c821afb3e09f3354e11f60927ee740d44cf)

---

### Test instructions

To test the "happy path" i.e. where everything is working as it should be, just use the Netlify link and connect to `https://debug.dhis2.org/dev`. Here you should be seeing the Superset dashboards load correctly. So this would be regression testing.

To test the changes implemented here, use the same netlify link and connect to another v42 test instance, for example `https://test.e2e.dhis2.org/analytics-dev`. Now check the "Enable embedded dashboards" checkbox [in system settings](https://test.e2e.dhis2.org/analytics-dev/apps/settings#/analytics). To see the implemented error view, just create an embedded dashboard and try to view it.

### Screenshot

Screenshot 2025-03-24 at 12 19 58

[DHIS2-19275]: https://dhis2.atlassian.net/browse/DHIS2-19275?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


jenniferarnesen:
I think we should consider a different solution. With this implementation, you are basically stuck and can't use the dashboard app at all, right? Instead of SystemSettingProvider returning a Notice component, it should just return the value of "hasConfigurationError" and then let the ViewDashboard decide how to handle it. It should still be possible to view other dashboards, create regular dashboards, etc. Just that the option to create embedded dashboards is removed and existing embedded dashboards could show an error, with a link to go to the dashboard start page.

jenniferarnesen (changes requested):
See other comment

src/components/SystemSettingsProvider.js
```diff
@@ -64,7 +68,7 @@ export const useIsSupersetSupported = () => {
     const {
         systemSettings: { embeddedDashboardsEnabled, supersetBaseUrl },
     } = useSystemSettings()
-    return embeddedDashboardsEnabled && !!supersetBaseUrl
+    return embeddedDashboardsEnabled && typeof supersetBaseUrl !== 'undefined'
```
jenniferarnesen: Here it is checking for `undefined` but on line 40, it is set to `null`. Just checking that that is what you intended.

HendrikThePendric: Yes this is exactly as intended. Either a string or `null` value are valid, and it being `undefined` makes it invalid. 

Would this make the intentions clearer though?
```js
return (
    embeddedDashboardsEnabled &&
    ((typeof supersetBaseUrl === 'string' && supersetBaseUrl.length > 0) ||
        supersetBaseUrl === null)
)
```

jenniferarnesen: I like your suggestion better, but I still don't fully understand how 'null' is a valid value, since it still isn't clear reading the code what that usecase is. Could we create a constant for 'null'? That would clarify what it is? 

I guess it means that the request wasn't 404 but there is something else currently wrong with the gateway, right?

jenniferarnesen: `SERVICE_IS_CURRENTLY_DOWN` or something like that.

HendrikThePendric: I'm not sure if this source is super reliable but this is basically how I've always understood the difference between `null` and `undefined` in JS: 
> Although `undefined` and `null` have some functional overlap, they have different purposes. In the strictest sense, `null` represents a value intentionally defined as "blank", and `undefined` represents a lack of any assigned value.

[source](https://web.dev/learn/javascript/data-types/null-undefined#comparison)

And IMO this is exactly what we're doing here.

Despite of this, I am not completely against adding a constant for this, I guess it could be a pragmatic solution. But it does also come with some downsides in my opinion. Specifically, what do we then do with `useSupersetBaseUrl`. This value is supposed to return a valid URL, and a string like `SERVICE_IS_CURRENTLY_DOWN` is not a valid URL. So then we end up writing an implementation of that hook which looks a bit like this:
```js
export const useSupersetBaseUrl = () => {
    const {
        systemSettings: { supersetBaseUrl },
    } = useSystemSettings()
    return supersetBaseUrl && supersetBaseUrl  !== SERVICE_DOWN ? supersetBaseUrl : null
}
```

And then I wonder, are we not just creating confusion elsewhere?

jenniferarnesen: I was thinking the constant would be set to null. So really it's just an explainer to us reading the code:
`const SERVICE_DOWN = null`

jenniferarnesen: But actually, I think it's fine with the earlier changes you suggested.

src/pages/view/SupersetDashboard.js
```diff
@@ -36,13 +36,26 @@ const reducer = (state, action) => {
     }
 }
 
+class SupersetGatewayUnavailableError extends Error {
+    constructor() {
+        super(
+            i18n.t(
+                'Could not load dashboard content because Superset Gateway is unavailable'
```
jenniferarnesen: I think there was a suggestion for a different error message in the feature channel.

jenniferarnesen (approved):

jenniferarnesen:
I've tested both the happy path and the error path and the implementation looks good. Testing approved.

PR #3285, opened 2025-06-20
feat: slideshow autoplay (DHIS2-18907)

jenniferarnesen:
Implements [DHIS2-18907](https://dhis2.atlassian.net/browse/DHIS2-18907)

Javascript timeout (setTimeout) is used to trigger the next slide in the slideshow.

### Description

* msPerSlide is stored in userDataStore and is used as the default
* User can select different msPerSlide (5 seconds is included for testing only)
* Manually clicking next or previous resets the timer
* Clicking pause pauses the timer
* Changing msPerSlide resets the timer to the new value

---

### Quality checklist

- [x] Docs added
- [ ] d2-ci dependency replaced (requires https://github.com/dhis2/analytics/pull/XXX) N/A
- [x] Small screen tested
- [x] Offline tested
- [x] Tester approved (@HendrikThePendric )

---

### Testing areas

- [x] Multiple filters applied to the dashboard
- [x] Offline
- [x] When only 1 fullscreen-enabled item is on the dashboard

---

[DHIS2-18907]: https://dhis2.atlassian.net/browse/DHIS2-18907?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/reducers/slideshow.js
```diff
@@ -1,8 +1,16 @@
 export const SET_SLIDESHOW = 'SET_SLIDESHOW'
+export const EXIT_SLIDESHOW = 'EXIT_SLIDESHOW'
 
-export default (state = null, action) => {
+const DEFAULT_SLIDESHOW_SETTINGS = {
+    firstItemIndex: null,
+    startPlaying: null,
+}
```
HendrikThePendric: Perhaps reconsider renaming this from "settings" so "state"?

src/actions/slideshow.js
```diff
@@ -1,6 +1,10 @@
-import { SET_SLIDESHOW } from '../reducers/slideshow.js'
+import { SET_SLIDESHOW, EXIT_SLIDESHOW } from '../reducers/slideshow.js'
 
-export const acSetSlideshow = (isSlideshow) => ({
+export const acSetSlideshow = (slideshowSettings) => ({
```
HendrikThePendric: See my other comment about renaming settings to sth else. Settings suggests to me these are config options.

src/pages/view/slideshow/useSlideshowAutoplay.js
```diff
@@ -0,0 +1,168 @@
...
+    },
+    [TWO_MINUTES]: {
+        label: i18n.t('2 minutes per slide'),
+        ms: 120000,
+    },
+})
+
+const KEY_SLIDESHOW_MS_PER_SLIDE = 'slideshowMsPerSlide'
+
+const timingOptions = getTimingOptions()
```
HendrikThePendric, line 39: I guess the `getTimingOptionsI` is there to avoid possible issues with the i18n string being in the wrong locale. I'm not sure this is still required these days, because I think the app-shell ensures the locale is correctly initialised before the app's modules are imported. But supposing that this is not the case, then you should not call the `getTimingOptions` function here in the module scope, but in the hook's function scope. The way it has been written now, it will effectively be the same as simply declaring the object here.

src/pages/view/slideshow/useSlideshowAutoplay.js
```diff
@@ -0,0 +1,168 @@
...
+                ms = parseInt(storedMsPerSlide) || DEFAULT_MS_PER_SLIDE
+            } catch (e) {
+                console.warn('Error fetching slideshow settings', e)
+                ms = DEFAULT_MS_PER_SLIDE
+            }
+
+            setMsPerSlide(ms)
+            setSlideMsRemaining(ms)
+        }
+        fetchMsPerSlide()
```
HendrikThePendric: Consider using `useCallback` to declare `fetchMsPerSlide` and calling it in the `useEffect` for improved readability.

src/pages/view/slideshow/useSlideshowAutoplay.js
```diff
@@ -0,0 +1,168 @@
...
+                    Math.max(slideMsRemainingRef.current - elapsed, 0)
+                )
+            }
+        }
+
+        prevNextItemRef.current = nextItem
+        prevMsPerSlideRef.current = msPerSlide
+
+        return () => timeoutRef.current && clearTimeout(timeoutRef.current)
+    }, [isPlaying, msPerSlide, nextItem])
```
HendrikThePendric: It took me a while before I fully understood the logic within this effect hook. It would be good if it could be somehow changed a bit so that it is easier to wrap your head around. 

I'll share some thoughts below:
- The fact that `nextItem` is a function and we are then also checking if the function changes like so `const nextItemChanged = nextItem !== prevNextItemRef.current` is kind of sending mixed messages. I guess it works because the function is being recreated at the right times, but I'm not sure this is a great solution going forward. If something happens and the nextItem function is recreated more or less frequently, then this code will break. Ideally we should respond to changes in of an integer, for example `slideshowItemIndex` is also being returned from the `useSlideshow` hook. And it would be possible to rewrite `nextItem` and `prevItem` to no longer have the `itemIndex` dependency.
- Essentially, why this is hard to reason about, is that we're responding to various variable-changes here, and then within the effect we are figuring out what it was that actually changed. Perhaps it is not possible to split this up, but it would be a lot more readable if there were three separate effects that respond to distinct variable-changes.

Perhaps a bit drastic, but you could consider not having this `useSlideshowAutoplay` hook, and instead introduce a class called sth like `SlideshowPlayer`. This could have the necessary internal mutable state and store the timeouts that are currently in refs. I would expose some methods like `nextItem`, `prevItem`, `toggleAutoplay`, `setMsPerSlide`. And then in `useSlideShow` you could do sth like this:

```js
const [itemIndex, setItemIndex] = useState(null)
const onItemIndexChange = useCallback((index) => {
    setItemIndex(index)
})
const [slideshowPlayer] = useState(() => new SlideshowPlayer(onItemIndexChange))
```

I none of my suggestion are feasible, or worth the effort, then perhaps just adding some comments to clarify what is going on will suffice

HendrikThePendric (commented):
I've left a few comment.

src/pages/view/slideshow/useSlideshowAutoplay.js
jenniferarnesen: I tried just using an object and it doesn't work, unfortunately. The translations only display if we use a function (you are right though that the function has to be run within the component)

src/pages/view/slideshow/useSlideshowAutoplay.js
jenniferarnesen: I'm not sure how much we would gain by doing this. The useCallback would have to be declared with an async function. In turn, useEffect would still have to define an async function to call the callback. But maybe I'm missing something?  For example:

```
const fetchMsPerSlide = useCallback(async () => {
      let ms = DEFAULT_MS_PER_SLIDE
      try {
          const storedMsPerSlide = await apiGetUserDataStoreValue(
              KEY_SLIDESHOW_MS_PER_SLIDE,
              DEFAULT_MS_PER_SLIDE,
              dataEngine
          )
          const intParsedMsPerSlide = parseInt(storedMsPerSlide)
          if (
              !isNaN(intParsedMsPerSlide) &&
              Object.values(timingOptions).some(
                  (option) => option.ms === intParsedMsPerSlide
              )
          ) {
              ms = intParsedMsPerSlide
          }
      } catch (e) {
          console.warn('Error fetching slideshow state', e)
      }

      return ms
}, [dataEngine])

useEffect(() => {
      const fetchit = async () => {
          const ms = await fetchMsPerSlide()
          setMsPerSlide(ms)
          setSlideMsRemaining(ms)
      }
      fetchit()
}, [fetchMsPerSlide])
```

src/pages/view/slideshow/useSlideshowAutoplay.js
jenniferarnesen: @HendrikThePendric I'm wondering if you would want to have a shot at this? I'm kind of at the point of not being able to see the forest for the trees on this one.

jenniferarnesen: > If something happens and the nextItem function is recreated more or less frequently

I'm not sure under what circumstances this would happen.

I do agree that the effect is hard to reason about.

src/pages/view/slideshow/useSlideshowAutoplay.js
HendrikThePendric: OK, good to know. When calling it in the component, you could even consider wrapping it in a `useMemo`, but I guess there's no major performance penalty in calling `i18n.t()` for a bunch of strings repeatedly. 

Quite interesting though, how just using an object didn't work, I've been at DHIS2 for about 8 years now and seem to remember different stages of localisation:
1. Way back we had something which was part of `d2`
2. Then we moved to `i18next` and we had this restriction of only calling `i18n.t()` in code that runs after the component has mounted (not in the module scope)
3. After that, I am 99% sure that there was a phase we we could forget about that aforementioned restriction because the app-shell was lazy-loading the component. It was possible to call `i18n.t()` in the module scope because of this. I've had a look and can see that code is still in place, see [here](https://github.com/dhis2/app-platform/blob/master/shell/src/App.jsx#L8-L11) and [here](https://github.com/dhis2/app-platform/blob/master/shell/src/App.jsx#L60-L70)
4. And then now it seems we've entered a new era again, where we have to take the original restriction into account again? Perhaps because of the move from CRA to Vite?

I wonder if this is perhaps something we should double check with team-extensibility? Or shall we just accept it for what it is?

src/pages/view/slideshow/useSlideshowAutoplay.js
HendrikThePendric: Good point, I overlooked that you were calling `setMsPerSlide` and `setSlideMsRemaining` in function body of `fetchMsPerSlide`. I see you have refactored the code already though, and I do slightly prefer this because now `fetchMsPerSlide` just does what it says on the tin.

src/pages/view/slideshow/useSlideshowAutoplay.js
HendrikThePendric: > If something happens and the nextItem function is recreated more or less frequently

Instead of saying "when something happens" I should have probably said something more specific like "when someone makes changes to the codebase in the future". For example, in `useSlideShow.js`:

```js
    const nextItem = useCallback(() => {
        if (itemIndex === sortedItems.length - 1) {
            setItemIndex(0)
        } else {
            setItemIndex(itemIndex + 1)
        }
    }, [itemIndex, sortedItems])
 ```
 
 It could be rewritten like so:
 ```js
    const nextItem = useCallback(
        () =>
            setItemIndex((currIndex) =>
                currIndex === sortedItems.length - 1 ? 0 : currIndex + 1
            ),
        [sortedItems]
    )
```
    
And now it doesn't get recreated nearly as often and we'd have a bug.

src/pages/view/slideshow/useSlideshowAutoplay.js
jenniferarnesen: > After that, I am 99% sure that there was a phase we we could forget about that aforementioned restriction because the app-shell was lazy-loading the component.

I remember people saying this, but i never saw any proof myself. We had already switched all the i18n calls to happen within the component function.

jenniferarnesen: I agree this needs to be looked at but I don't think the performance impact on the slideshow would be noticeable.

src/pages/view/slideshow/useSlideshowAutoplay.js
jenniferarnesen: @HendrikThePendric I've made some of the changes you suggested. I think the code is better now, but let me know.

src/pages/view/slideshow/useSlideshowAutoplay.js
HendrikThePendric: OK, I'll report it to team-extensibility. For now it's not relevant, we keep the code as-is

src/pages/view/slideshow/useSlideshowAutoplay.js
HendrikThePendric: Yeah, this is pretty easy to follow now, thanks 🚀

HendrikThePendric (commented):
The code is looking good and the feature seems to work well. The only minor thing I noticed is that auto-play is enabled by default, and I suspect it should probably not be.

I also some some rendering issues on map visualisations, but I can't imagine this is related to the current PR.

I will probably have to re-approve because I know you're still working on some things, so I'll just set the status to "Comment". But I won't be requestingr any more changes.

HendrikThePendric (dismissed):
Approved because the auto-play auto-start behaviour was according to the specs

HendrikThePendric (dismissed):

src/pages/view/slideshow/CountdownBar.jsx
```diff
@@ -0,0 +1,53 @@
...
+            progressEl?.animate(
+                [{ inlineSize: `${width}px` }, { inlineSize: '0px' }],
+                {
+                    duration: slideMsRemaining,
+                    fill: 'forwards',
+                    easing: 'linear',
+                }
+            )
+        }
+    }, [progressEl, slideMsRemaining, width, isPlaying])
```
HendrikThePendric: I haven't tested this, but by just looking at the code, I'd say we're missing something here: should't we also be able to pause a pending animation?

By [the look of it](https://developer.mozilla.org/en-US/docs/Web/API/Web_Animations_API/Using_the_Web_Animations_API#pausing_and_playing_animations), adding this feature would be fairly straightforward. You would need to store a ref to an animation and call `animationRef?.pause()`

src/pages/view/slideshow/CountdownBar.jsx
```diff
@@ -0,0 +1,53 @@
...
+
+    return (
+        <>
+            <div
+                ref={progressElRef}
+                data-test="countdown"
+                className={cx(styles.progress, { [styles.paused]: !isPlaying })}
+                style={{
+                    inlineSize: `${width}px`,
+                }}
```
HendrikThePendric, line 21: I don't fully understand this. Shouldn't we either use the React API to control the width, OR use the native web API. Seems like now we're doing both...

HendrikThePendric (commented):
I've left a few comments/questions, all reg. the new `CountdownBar`. I feel like we should either control the `width` and `animation-duration` via React state, or we should use the `Element.animate()` API. We seem to have landed on kind of a dual approach and I'm not sure if that is a good idea....

I'll also do some testing now and will report back in a comment.

HendrikThePendric:
I've just tested the feature using the Netlify link and it working well.

src/pages/view/slideshow/AutoplayControls.jsx
```diff
@@ -58,11 +60,20 @@ const AutoplayControls = ({ nextItem, itemIndex }) => {
...
                 data-test="slideshow-autoplay-play-pause-button"
             >
+                {slideMsRemaining !== null && (
+                    <CountdownBar
+                        key={itemIndex}
+                        isPlaying={isPlaying}
+                        msPerSlide={msPerSlide}
+                        slideMsRemaining={slideMsRemaining}
+                    />
+                )}
```
HendrikThePendric: I've just come to the following realisation:
We are not actually showing the `CountdownBar` when `isPlaying` is `false`. If we were to also not render it when this is the case, we could seriously simplify the countdown bar: it would just need to have a computed width and animation-duration. And apart from that it could just have a CSS keyframe animation that animates its width to zero.

HendrikThePendric: Also, why is the `key` needed?

src/pages/view/slideshow/CountdownBar.jsx
HendrikThePendric: I believe this comment no longer makes sense, given my suggestions to simply not render the component when `isPlaying` is `false`.

HendrikThePendric (commented):
I re-reviewed, because I reconsidered the implementation and think it can be simplified.

src/pages/view/slideshow/AutoplayControls.jsx
HendrikThePendric: Actually, ignore my question about `key={itemIndex}`. This will force a remount of the component and is actually an important "ingredient" that we must keep if we want to simplify things as I suggested.

src/pages/view/slideshow/CountdownBar.jsx
jenniferarnesen: @HendrikThePendric Sorry, I don't quite follow.

src/pages/view/slideshow/AutoplayControls.jsx
jenniferarnesen: Removed `key`. It should have been `itemIndex`.

As far as your other suggestion, I tried last week to use css animation and keyframes directly, to no avail. I'm not sure its actually possible to change the value and have it restart the animation (e.g. while changing the slide or the slide duration while the slideshow is playing).

HendrikThePendric (dismissed):
Looks great! This is a very neat feature 😻

src/pages/view/slideshow/AutoplayControls.jsx
jenniferarnesen: Just to conclude, key is now used once again to enforce a remount of the component when either the itemIndex or the msPerSlide changes.

HendrikThePendric (approved):

src/pages/view/slideshow/CountdownBar.jsx
jenniferarnesen: Resolving as this implementation was simplified.

PR #3335, opened 2026-07-16
fix: hide "Open in app" for plugins without an app entrypoint (DHIS2-21739)

karolinelien:
Implements [DHIS2-21739](https://dhis2.atlassian.net/browse/DHIS2-21739)

### Description

Plugin-only apps (installed without an app entrypoint in `d2.config.js`, e.g. Capture Growth Chart) are served with a `pluginLaunchUrl` but no `launchUrl` — there is no standalone app to open. Despite this, the dashboard item's ⋯ menu still showed an **"Open in {app}"** option. Clicking it navigated to a broken relative URL (`.../dhis-web-dashboard/undefinedundefined`) and 404'd, because the app URL was built by concatenating `` `${appDetails.launchUrl}${newState.appUrl}` `` where both parts were `undefined`.

This PR:

- Only exposes the "Open in {app}" option when the app actually has a launch URL (`openInAppEnabled = allowVisOpenInApp && !!appUrl`). The ⋯ menu is still shown for such plugins when "View fullscreen" is available.
- Stops the reducer from stringifying `undefined` when there is no `launchUrl`, so `appUrl` is `undefined` (not `"undefinedundefined"`) for entrypoint-less plugins.

AI Assisted.

---

### Quality checklist

- [x] Docs added — _N/A_
- [x] d2-ci dependency replaced (requires https://github.com/dhis2/analytics/pull/XXX) — _N/A_
- [x] Small screen tested
- [x] Offline tested
- [x] Tester approved (@HendrikThePendric)

---

### ToDos

N/A

---

### Known issues

N/A

---

### Screenshots

_supporting text_

[DHIS2-21739]: https://dhis2.atlassian.net/browse/DHIS2-21739?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/Item/AppItem/Item.jsx
+++ b/src/components/Item/AppItem/Item.jsx
@@ -63,7 +63,11 @@ const AppItem = ({
         (state, newState) => ({
             ...state,
             ...newState,
-            appUrl: `${appDetails?.launchUrl}${newState.appUrl}`,
+            // plugins without an app entrypoint have no launchUrl, so there is
+            // nothing to open in a standalone app (DHIS2-21739)
+            appUrl: appDetails?.launchUrl
+                ? `${appDetails.launchUrl}${newState.appUrl ?? ''}`
+                : undefined,
         }),
         {
             itemTitle: appDetails?.name,
--- a/src/components/Item/AppItem/Item.jsx
+++ b/src/components/Item/AppItem/Item.jsx
@@ -30,6 +30,7 @@ import ItemHeader from '../ItemHeader/ItemHeader.jsx'
 import MissingPluginMessage from '../ItemMessage/MissingPluginMessage.jsx'
 import { getIframeSrc } from './getIframeSrc.js'
 import { ItemContextMenu } from './ItemContextMenu.jsx'
+import { createItemDetailsReducer } from './itemDetailsReducer.js'
 import styles from './styles/AppItem.module.css'
 
 const AppItem = ({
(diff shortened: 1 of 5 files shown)


src/components/Item/AppItem/Item.jsx
```diff
@@ -162,7 +159,6 @@ const AppItem = ({
             isViewMode(dashboardMode) &&
             !isSlideshowView ? (
                 <ItemContextMenu
-                    item={item}
```
HendrikThePendric, line 165: This prop was removed and this is correct: `ItemContextMenu` does not have an `item` prop.

src/components/Item/AppItem/__tests__/itemDetailsReducer.spec.js
```diff
@@ -0,0 +1,30 @@
+import { createItemDetailsReducer } from '../itemDetailsReducer.js'
+
+// DHIS2-21739: a plugin-only app has a pluginLaunchUrl but no launchUrl, so
+// there is no standalone app to open. The reducer used to build the href as
+// `${launchUrl}${appUrl}`, yielding the literal string "undefinedundefined".
```
HendrikThePendric:
```suggestion
```
I think this comment is a bit redundant, the tests actually describe this behaviour and you already reference the issue number and desired behaviour in other places

HendrikThePendric (dismissed):
Thanks for this contribution @karolinelien 🥇 

When I first looked at the code I was going to suggest changing things around and keeping the reducer inline in `Item.jsx` rather than extracting `createItemDetailsReducer`. This would have made the code a bit easier to reason about. 

But on closer inspection I think the implementation you went with is the right call. In the test for `Item.jsx` we assert that no "Open in app" menu item is showing when there is no `appUrl`. But this isn't the full story: it's good to have a test that clearly describes how this `appUrl` becomes undefined.

I left an optional suggestion in the code and an observation.

PR #3337, opened 2026-08-03
test: guard against clicking disabled slideshow button (DHIS2-21933)

karolinelien:
https://dhis2.atlassian.net/browse/DHIS2-21933

### Description

Small hardening fix for `cypress/integration/slideshow.cy.js`:

Swaps the text-based `cy.get('button').contains('Slideshow')` selector for the button's
existing `data-test="enter-slideshow-button"` hook, and adds a `.should('not.be.disabled')`
guard before clicking it. The button can be legitimately disabled offline/embedded/before
items load, and `.should('not.be.disabled')` makes Cypress retry until it's enabled instead
of risking `realClick()` silently no-op'ing on a disabled button.

The two `cy.wait(11000)` calls in the autoplay test are kept as-is: they assert the slide has
*not yet* changed before the real 10s autoplay timer fires, which a retrying assertion
wouldn't correctly verify, so a fixed wait is the right tool there.

No production code (`src/`) changed. Not a fix for the broader Cypress suite flakiness —
just a small, targeted improvement to this one selector.

---

### Quality checklist

- [ ] Tested with and without global shell — N/A, no shell-affecting change
- [ ] Docs added — N/A
- [ ] d2-ci dependency replaced (requires https://github.com/dhis2/analytics/pull/XXX) — N/A
- [ ] Small screen tested — N/A
- [ ] Offline tested — N/A
- [ ] Tester approved (name)

---

### ToDos

- [ ] _todo_

---

### Known issues

- [ ] _issue_

---

### Screenshots

_supporting text_

AI Assisted


HendrikThePendric (approved):
As discussed with @karolinelien, while I do not expect this to help "de-flaking" very much,  this is still a valid improvement. It's better to wait for the button to be enabled before clicking. 

Perhaps it would be good to go with a slightly less ambitious PR title ;-)

PR #3344, opened 2026-09-17
feat: add support for EVER plugin in 43+ [DHIS2-21962]

edoardo:
Implements [DHIS2-21962](https://dhis2.atlassian.net/browse/DHIS2-21962)

### Description

The new EVER app replaces Line Listing as the editor and dashboard plugin for event visualizations, and it handles both `LINE_LIST` and `PIVOT_TABLE` types.

The EVER app is only available from v43, which means we need to still use the LL app in supported previous versions.
This is done using version toggles around the code, as setting a minDHIS2Version to 2.43 in the dashboard app means we loose the ability to ship fixes for older versions.

Key differences:
- **below 2.43** — LL plugin, search result group titled "Line lists", dashboard fetches the visualization and shows its own "filters not applied" notice
- **2.43 and above** — EVER plugin, search result group titled "Event visualizations" and using different icons based on the visType, no fetch and no "filters not applied" notice since the plugin takes care of all that

For this to work, many parts needed to access `apiVersion`.

New specs have been added for parts previously untested.

### Notes on related fixes required

- **Backend:** `dashboards/search` currently only returns event visualization of type `LINE_LIST`, it should return `PIVOT_TABLE` ones too in 43+ (see )
- **EVER app key:** with this PR the dashboard expects `individual-data-visualizer` as app key; this needs to be finalised and changed in EVER's `d2.config.js`
- **EVER plugin props:** for "standalone" plugins, dashboard now sends`visualizationId` instead of a `visualization` object, so the EVER plugin props needs to be changed accordingly

---

### Known issues

1. after discussing with @HendrikThePendric about the need for backend version checks, it emerged that the work on "tagging" a event visualization with the app that generated/saved it has an impact on the solution adopted in the PR.
The version check can be replaced with a different check that takes into account the app presence and the app that is "stamped" in the visualization object.
See the relevant ticket https://dhis2.atlassian.net/browse/DHIS2-22143

---

### Quality checklist

- [ ] Tested with and without global shell
- [ ] Cypress and/or Jest tests added/updated
- [ ] Docs added N/A
- [ ] d2-ci dependency replaced (requires https://github.com/dhis2/analytics/pull/XXX) N/A
- [ ] Small screen tested
- [ ] Offline tested
- [ ] Tester approved (name)

---

### Screenshots

Search dropdown now shows a "Event visualizations" section in place of "Line lists":
Screenshot 2026-09-17 at 14 33 49

LL visualization rendered with the EVER plugin:
Screenshot 2026-09-17 at 11 24 25

[DHIS2-21962]: https://dhis2.atlassian.net/browse/DHIS2-21962?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/modules/isAppVersionCompatible.js
```diff
@@ -1,4 +1,9 @@
+// Minimum DHIS2 api version that returns PIVOT_TABLE event visualizations and bundles the EVER app
+export const MIN_API_VERSION_FOR_EVER = 43
```
HendrikThePendric: Not sure about this part. I was hoping we could "outsource this part to the app-hub / d2.config file". My thoughts were:
1. The EVER app declares its own min version and this prevents it from being installed onto an unsuitable backend
2. So here in the dashboard-app we can assume that when the app is found, the backend version must be OK

HendrikThePendric (commented):
The quality of the code and test coverage is looking great. I left one short comment in the code but it is actually quite a fundamental point. Do we need to concern ourselves with API versions etc? I can imagine that for some older apps/plugins it is relevant to concern ourselves with app-versions and/or backend versions. But for this new app/plugin I was hoping we could just assume all is well if the app was allowed to get installed....

src/modules/isAppVersionCompatible.js
HendrikThePendric: 🤔 The only exception I just came up with is when neither LL nor EVER is installed and we prompt the user to install an app (or get the sys admin to do so I guess). In this case the backend version is the only way to figure out which app is the best fit for the current instance.

But actually the best you could do is:
1. If backend is below the EVER release version, the prompt for LL
2. If backend supports EVER, then it's actually possible to install LL or EVER, so the prompt should maybe suggest both

HendrikThePendric: And if both LL and EVER are installed on an instance we agreed we'd use EVER in the dashboard-app... But still you wouldn't need a version for it....

edoardo: The `MIN_API_VERSION_FOR_EVER` is still needed for the parts where we want to verify the current version to decide what to show, for example in the search results list where we want to show "Event visualizations" instead of "Line lists".
This doesn't depend on the fact that the EVER app is installed, but solely on the backend version that returns PT visualizations (43+).

This actually made me think that the constant's name is confusing and perhaps is what triggered you comment.
I now dropped the constant and use the plain api version directly, since it's also unlikely we need to update the constant with a different version.

About point 2: relying on installed app to choose what to do, we don't do that for LL either.
It's always been based on api version and now is not different.
If we are on a 43+ instance, the list will contain PT visualizations regardless of wether the EVER app is installed.
There is already a mechanism in place that tells the user to install a certain version of an app to open a certain visualization.

Happy to discuss this further.

src/modules/itemTypes.js
```diff
@@ -65,12 +70,12 @@ export const itemTypeMap = {
...
         dataStatisticsName: 'VISUALIZATION_VIEW',
         propName: 'visualization',
-        pluralTitle: i18n.t('Visualizations'),
+        pluralTitle: () => i18n.t('Visualizations'),
         domainType: DOMAIN_TYPE_AGGREGATE,
         isVisualizationType: true,
         appUrl: ({ id }) => `dhis-web-data-visualizer/#/${id}`,
-        appName: 'Data Visualizer',
-        appKey: 'data-visualizer',
+        appName: () => 'Data Visualizer',
```
HendrikThePendric: i18n

HendrikThePendric (approved):
Approving this PR with the following in mind:
- The current backend version based logic to determine which app/plugin to use will be replaced in the future
- The fact that `appName` is a hardcoded English string, not a localised one, is intentional

